PR #30, opened 2020-05-19
Updates authorized_key.py to be used with FIDO2 security keys.

fbettag:
Last try had a space at the end for some weird reason.


fbettag:
The list is now current with all openssh capable key types as outlined here:
https://man.openbsd.org/sshd#AUTHORIZED_KEYS_FILE_FORMAT

PR #31, opened 2020-05-24
Synchronize plugin: Do not turn an SSH path into an absolute path

AlbertJP:
##### SUMMARY
SSH paths of the form user@host:/remote/path can be used as either src or dest with rsync.
However, the path is turned into /local/path/user@host:/remote/path, which is not a valid syntax for rsync. This commit fixes that problem.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
synchronize action plugin

##### ADDITIONAL INFORMATION
As the synchronize module already uses the presence of ':' to detect a source/destination over SSH, this commit updates the action plugin accordingly.

### DIFF
--- a/plugins/action/synchronize.py
+++ b/plugins/action/synchronize.py
@@ -33,7 +33,7 @@ class ActionModule(ActionBase):
     def _get_absolute_path(self, path):
         original_path = path
 
-        if path.startswith('rsync://') or path.startswith('/') or ':' in path:
+        if path.startswith(('rsync://', '/')) or ':' in path:
             return path
 
         if self._task._role is not None:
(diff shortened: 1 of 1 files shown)


AlbertJP:
Your unit tests appear to be broken, by the way, as every one of them fails with 'ERROR! Cannot meet dependency requirement 'ansible.netcommon:>=0.1.0' for collection community.general from source 'https://galaxy.ansible.com/api/'. Available versions before last requirement added: 0.0.1, 0.0.2'.

maxamillion:
recheck

plugins/action/synchronize.py
```diff
@@ -33,7 +33,7 @@ class ActionModule(ActionBase):
     def _get_absolute_path(self, path):
         original_path = path
 
-        if path.startswith('rsync://'):
+        if path.startswith('rsync://') or path.startswith('/') or ':' in path:
```
Akasurde, line 36:
```suggestion
        if path.startswith(('rsync://', '/')) or ':' in path:
```

PR #33, opened 2020-05-29
revert "mount: Check if src exists before mounted (ansible/ansible#61752)"

Akasurde:
##### SUMMARY

Partially revert "mount: Check if src exists before mounted (ansible/ansible#61752)"

This reverts part of ansible commit 72023d7462e78635264fd12bfdb23894b4163cba.

The immediate reason is that it breaks mounts where src is not a path.
Examples of such mounts are network-based filesystems such as nfs, cifs,
glusterfs, ceph, virtual filesystems such as tmpfs or overlayfs, and
also UUID-based mounts. It is too hard to come with an exhaustive list,
especially if we take non-Linux systems into account, so don't even try.

Additionally, it did not really fix the issue (ansible/ansible#59183) that
it intended to fix, because the mount could fail but leave a non-working
fstab entry for reasons other than non-existing src path.

Restore fstab and remove the mount point after a failed mount

Add a reminder that not only devices can be mounted

Fixes: ansible/ansible#65855
Fixes: ansible/ansible#67588
Fixes: ansible/ansible#67966

Signed-off-by: Alexander E. Patrakov <patrakov@gmail.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/mount.py
tests/integration/targets/mount/aliases

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -29,7 +29,7 @@ options:
     aliases: [ name ]
   src:
     description:
-      - Device to be mounted on I(path).
+      - Device (or NFS volume, or something else) to be mounted on I(path).
       - Required when I(state) set to C(present) or C(mounted).
     type: path
   fstype:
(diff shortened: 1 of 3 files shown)


gundalow:
Looks like this contains some other fixed. Are they needed to allow the integration tests to pass?

Akasurde:
> Looks like this contains some other fixed. Are they needed to allow the integration tests to pass?

Other changes are refactoring changes so I think we are good on that side.

PR #67, opened 2020-07-06
patch - add option to ignore whitespace changes

maxaudron:
##### SUMMARY
Add the option to ignore whitespace changes between patch and input.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
patch

### DIFF
--- a/plugins/modules/patch.py
+++ b/plugins/modules/patch.py
@@ -74,6 +74,11 @@ options:
       - If set to C(no), C(patch) will replace CRLF in C(src) files on POSIX.
     type: bool
     default: no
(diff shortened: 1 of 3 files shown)


maxamillion:
looks like sanity tests are upset about line length, but other than that the patch looks good to me 

```
ERROR: Found 1 pep8 issue(s) which need to be resolved:
ERROR: plugins/modules/patch.py:210:161: E501: line too long (164 > 160 characters)
```

Would you mind adding a test case or two to `tests/integration/targets/patch/tasks/main.yml` for good measure against regressions in the future? Thanks!

PR #99, opened 2020-09-25
at: add support for AIX

dberg1:
### SUMMARY

The `at` module does not work correctly on AIX.
Trying to remove a job fails with a syntax error.
Setting the `unique` attribute to `yes` also fails.
AIX uses `-lv` options to cat a job (instead of `-c` for Linux) and `-r` to remove a job.
Linux supports both `-r` and `-d`, so use `-r` since it is what POSIX says.
See https://www.ibm.com/support/knowledgecenter/ssw_aix_72/a_commands/at.html and https://pubs.opengroup.org/onlinepubs/9699919799/utilities/at.html
Other systems like SunOS or the BSDs also appear to use `-r` to remove a job.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

at

##### ADDITIONAL INFORMATION

To recreate the issue, use the following playbook:
```
- hosts: aix
  gather_facts: no
  tasks:
  - at:
      command: "ls -d / > /dev/null"
      state: absent
```
This fails with:
```
fatal: [aix-host1]: FAILED! => {
    "changed": false,
    "cmd": "/usr/bin/at -c root.1599830992c0.a",
    "invocation": {
        "module_args": {
            "command": "ls -d / > /dev/null",
            "count": null,
            "script_file": null,
            "state": "absent",
            "unique": false,
            "units": null
        }
    },
    "msg": "at: syntax error",
    "rc": 1,
    "stderr": "at: syntax error\n",
    "stderr_lines": [
        "at: syntax error"
    ],
    "stdout": "",
    "stdout_lines": []
}
```

### DIFF
--- a/plugins/modules/at.py
+++ b/plugins/modules/at.py
@@ -74,6 +74,7 @@ EXAMPLES = r'''
 '''
 
 import os
+import platform
 import tempfile
 
 from ansible.module_utils.basic import AnsibleModule
@@ -89,7 +90,7 @@ def add_job(module, result, at_cmd, count, units, command, script_file):
 
 def delete_job(module, result, at_cmd, command, script_file):
     for matching_job in get_matching_jobs(module, at_cmd, script_file):
-        at_command = "%s -d %s" % (at_cmd, matching_job)
+        at_command = "%s -r %s" % (at_cmd, matching_job)
         rc, out, err = module.run_command(at_command, check_rc=True)
         result['changed'] = True
     if command:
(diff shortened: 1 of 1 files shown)


maxamillion:
I don't have a way to test this myself because I don't have access to an AIX system but the code change looks good, thank you for the contribution!

maxamillion (approved):

czi:
Thank you for this.  Issue 62153 / bugfix 62154 got ignored by Redhat and closed with the move to ansible-posix.  I'm glad to see this finally merged!

PR #117, opened 2020-12-07
Fix module in order to don't show "changed" status when running ansib…

Bilanda:
…le playbook with authorized_key module if keys are not different

##### SUMMARY

When running ansible playbook with this module, even if keys are not different, it shows "changed" status instead of "ok".

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

authorized_key.py

##### ADDITIONAL INFORMATION

To reproduce :
Run a playbook using this module without fix
Run it again
See it shows "changed", but nothing changed
Run it with fix
No "changed" anymore

```paste below

```

### DIFF
--- a/plugins/modules/authorized_key.py
+++ b/plugins/modules/authorized_key.py
@@ -618,9 +618,12 @@ def enforce_state(module, params):
     # for 'exclusive', make sure keys are written in the order the new keys were
     if state == "present" and exclusive:
         to_remove = frozenset(existing_keys).difference(keys_to_exist)
-        for key in to_remove:
-            del existing_keys[key]
-            do_write = True
+        if len(to_remove) != 0:
+            for key in to_remove:
+                del existing_keys[key]
+                do_write = True
+        else:
+            do_write = False
 
     if do_write:
         filename = keyfile(module, user, do_write, path, manage_dir, follow)


aminvakil:
It does not show `changed: true` when nothing has been changed.
Could you please test again with this playbook? (I've just put `exclusive` because you have changed the part with `exclusive` option, but that does not make any difference, feel free to test without it as well)
 ```
- name: test authorized_keys
  hosts: localhost
  tasks:
      - name: Get github key and update the authorized_keys file
        authorized_key:
          user: username
          exclusive: yes
          key: "https://github.com/aminvakil.keys"
          state: present
```

**Comment by maxamillion** [comment]
Thank you for the contribution! Would you mind adding a test case or two in `tests/integration/targets/authorized_key/tasks/main.yml`?

**Comment by Andersson007** [comment]
@Bilanda , is there a way we can help?

we also need a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)

File: plugins/modules/authorized_key.py [hunk 1]
```diff
@@ -618,9 +618,12 @@ def enforce_state(module, params):
...
         to_remove = frozenset(existing_keys).difference(keys_to_exist)
-        for key in to_remove:
-            del existing_keys[key]
-            do_write = True
+        if len(to_remove) != 0:
+            for key in to_remove:
+                del existing_keys[key]
+                do_write = True
+        else:
+            do_write = False
```
quidame, line 626: hi @Bilanda thanks for your contribution

I think this could open another issue. What I mean is:
- `do_write` has been initialized to `False`
- now you override it to `False` again
- what happened between ?
I guess that if there is an issue somewhere, it could rather be found after line 594 and before line 615, not here

In this fix, `do_write` is ***forced*** to `False` if we want the key to be `present` and `exclusive`, and if there is no key to remove. So, what in the case there is nothing to remove but something to add (the key we want to be present and exclusive and that is not already there) ? We get `do_write=False`, and I'm not sure this is want we want.

There is something unclear for me here, or in the description, or both. As mentioned first by @aminvakil , this is related to the `exclusive` option. Could you please provide a sample play that produces the error without your change, and that doesn't produce the error with your change ? It would be the best starting point for adding integration tests :)

quidame: Looking at the logic in the code, `do_write` is initialized to `False`, and possibly overridden to `True` in a loop, that means it is not expected to be reset to `False`, neither in the loop itself, nor at any time, without breaking the logic: `do_write` is a kind of *no-return-switch*, that once toggled to the opposite of its initial value, never goes back. Or think of it as an ORed sequence of booleans: (false or false or false or true or false or false) is `true`.

Many modules make use of this logic for the `changed` variable that is returned in their results.

Bilanda: Hello @quidame, thanks a lot for your support.
As mentionned before, i tried with @aminvakil's playbook and i still have the issue.
I used this kind of task from my side where i have the problem : 

`- name: Install root SSH authorized keys
  authorized_key:
    state: present
    user: root
    manage_dir: yes
    exclusive: yes
    key: "{{ lookup('template', 'root/.ssh/authorized_keys.j2') }}"`

But whenever i run the playbook, even if there is absolutely no change in the deployed keys, the task return _changed_
I can run playbook in verbose mode if this can help.
Thanks again

Akasurde (changes requested):
Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example. Also, write tests for this change.

quidame:
As stated in my previous https://github.com/ansible-collections/ansible.posix/pull/117#discussion_r640090945, this PR doesn't fix anything.

1. I can't reproduce the issue with the mainline version of the module (i.e. in all my tests - 1 key online, or 2 keys in a local file - I still get `changed=True` when something is changed on the target, and `changed=False` when nothing changed). @Bilanda are you sure the version of the module producing the issue for you is the mainline one ? Do `ansible-doc authorized_key` and `ansible-doc ansible.posix.authorized_key` show the same path in the first line ? Or: what is the content of your authorized_keys.j2 template ?

2. I can easily reproduce a new issue with the PR's version of the module, and this playbook:
```yaml
- name: test authorized_keys
  hosts: all
  tasks:
    - name: Remove authorized_keys file
      file:
        path: "/home/{{ ansible_user }}/.ssh/authorized_keys"
        state: absent

    - name: Get github key and update the authorized_keys file
      authorized_key:
        key: "{{ lookup('file', 'files/authorized_keys') }}"
        user: "{{ ansible_user }}"
        exclusive: yes
        state: present
      register: authorized_keys
      failed_when: authorized_keys is not changed
```
As explained, since there is no key to remove from the `.ssh/authorized_keys` (because the file itself is absent), your construction makes that `do_write` is `False`, and the expected `.ssh/authorized_keys` file is NOT created !!!

Bilanda:
Hi, 
Indeed issue comes from the inside use of my lookup through the template file.
Verry sorry for mistaking myself. 
Thanks for your help. i'm closing the issue.

Issue #100, opened 2020-09-25
firewalld module should support management of port forwards


misilot:
##### SUMMARY

firewalld module should support management of port forwards that were created via the --add-forward-port option
```$ firewall-cmd --add-forward-port=port=500:proto=tcp:toport=3753```
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
firewalld module

##### ADDITIONAL INFORMATION

The documentation allows for creating and modifying "rich rules" for managing forwards, however forwards that are already in place on the system cannot be managed without using the command module. So you could possibly have competing forwards in place, and I am not sure which one would win.

PR #129, opened 2021-01-16
firewalld: Add support for firewalld port forwarding

evan-a-a:
##### SUMMARY

Implement support in firewalld for use of the port forward API. Previously, the only way to forward ports using ansible was with the rich rule syntax. This change adds the `port_forward` parameter to manage port forwards directly.  The naming scheme for the parameter list is the same as existing firewalld naming. I opted to use a subgroup to clearly delineate the options and support better validation of the protocol parameter. 

Fixes: ansible-collections/ansible.posix#100
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
firewalld

### DIFF
--- a/docs/ansible.posix.firewalld_module.rst
+++ b/docs/ansible.posix.firewalld_module.rst
@@ -176,6 +176,21 @@ Parameters
             <tr>
                 <td colspan="1">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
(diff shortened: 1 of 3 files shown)


maxamillion (approved):

maxamillion:
@evan-a-a Thank you for the contribution, this is greatly appreciated!

PR #142, opened 2021-02-22
selinux: update kernel boot params when disabling/re-enabling SELinux

WOnder93:
##### SUMMARY

The ability to disable SELinux from userspace based on the configuration
file is being deprecated in favor of the selinux=0 kernel boot
parameter. (Note that this affects only the "full" disable; switching
to/from permissive mode will work the same as before.)

Therefore, enhance the selinux module to try to set/unset the kernel
command-line parameter using grubby when enabling/disabling SELinux.

If the grubby package is not present on the system, the module will only
update the config file and report a warning. Note that even with the
runtime disable functionality removed, setting SELINUX=disabled in the
config file will lead to a system with no SELinux policy loaded, which
will behave in a very similar way as if SELinux was fully disabled, only
there could still be some minor performance impact, since the kernel
hooks will still be active.

More information:
https://lore.kernel.org/selinux/157836784986.560897.13893922675143903084.stgit@chester/
https://fedoraproject.org/wiki/Changes/Remove_Support_For_SELinux_Runtime_Disable

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
selinux module

### DIFF
--- a/changelogs/fragments/disable_selinux_via_kernel_cmdline.yml
+++ b/changelogs/fragments/disable_selinux_via_kernel_cmdline.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- selinux - optionally update kernel boot params when disabling/re-enabling SELinux (https://github.com/ansible-collections/ansible.posix/pull/142).
--- a/plugins/modules/selinux.py
+++ b/plugins/modules/selinux.py
@@ -28,6 +28,13 @@ options:
     required: true
     choices: [ disabled, enforcing, permissive ]
     type: str
+  update_kernel_param:
+    description:
+      - If set to I(true), will update also the kernel boot parameters when disabling/enabling SELinux.
+      - The C(grubby) tool must be present on the target system for this to work.
+    default: no
+    type: bool
+    version_added: '1.4.0'
   configfile:
     description:
       - The path to the SELinux configuration file, if non-standard.
@@ -97,6 +104,7 @@ except ImportError:
     HAS_SELINUX = False
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.common.process import get_bin_path
 from ansible.module_utils.facts.utils import get_file_lines
 
 
@@ -117,6 +125,34 @@ def get_config_policy(configfile):
         stateline = re.match(r'^SELINUXTYPE=.*$', line)
         if stateline:
             return line.split('=')[1].strip()
+
+
+def get_kernel_enabled(module, grubby_bin):
+    if grubby_bin is None:
+        module.fail_json(msg="'grubby' command not found on host",
+                         details="In order to update the kernel command line"
+                                 "enabled/disabled setting, the grubby package"
+                                 "needs to be present on the system.")
+
+    rc, stdout, stderr = module.run_command([grubby_bin, '--info=ALL'])
+    if rc != 0:
+        module.fail_json(msg="unable to run grubby")
+
+    all_enabled = True
+    all_disabled = True
+    for line in stdout.split('\n'):
+        match = re.match('^args="(.*)"$', line)
+        if match is None:
+            continue
+        args = match.group(1).split(' ')
+        if 'selinux=0' in args:
+            all_enabled = False
+        else:
+            all_disabled = False
+    if all_disabled == all_enabled:
+        # inconsistent config - return None to force update
+        return None
+    return all_enabled
 
 
 # setter subroutines
@@ -153,6 +189,17 @@ def set_state(module, state):
         module.fail_json(msg=msg)
 
 
+def set_kernel_enabled(module, grubby_bin, value):
+    rc, stdout, stderr = module.run_command([grubby_bin, '--update-kernel=ALL',
+                                             '--remove-args' if value else '--args',
+                                             'selinux=0'])
+    if rc != 0:
+        if value:
+            module.fail_json(msg='unable to remove selinux=0 from kernel config')
+        else:
+            module.fail_json(msg='unable to add selinux=0 to kernel config')
+
+
 def set_config_policy(module, policy, configfile):
     if not os.path.exists('/etc/selinux/%s/policy' % policy):
         module.fail_json(msg='Policy %s does not exist in /etc/selinux/' % policy)
@@ -183,6 +230,7 @@ def main():
             policy=dict(type='str'),
             state=dict(type='str', required=True, choices=['enforcing', 'permissive', 'disabled']),
             configfile=dict(type='str', default='/etc/selinux/config', aliases=['conf', 'file']),
+            update_kernel_param=dict(type='bool', default=False),
         ),
         supports_check_mode=True,
     )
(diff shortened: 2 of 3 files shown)


Akasurde:
recheck

plugins/modules/selinux.py
```diff
@@ -150,6 +174,17 @@ def set_state(module, state):
         module.fail_json(msg=msg)
 
 
+def set_kernel_enabled(module, value):
+    rc, stdout, stderr = module.run_command(['grubby', '--update-kernel=ALL',
+                                             '--args' if value else '--remove-args',
+                                             'selinux=0'])
```
maxamillion, line 180: maybe I'm misunderstand, but this looks like it's disabling selinux instead of enabling

WOnder93: Yes, you're right. I got it backwards both here and in the function call, so it just happened to work correctly. I'll update the code to make more sense...

changelogs/fragments/disable_selinux_via_kernel_cmdline.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- update kernel boot params when disabling/re-enabling SELinux
```
Akasurde, line 2:
```suggestion
- selinux - update kernel boot params when disabling/re-enabling SELinux (https://github.com/ansible-collections/ansible.posix/pull/142).
```

plugins/modules/selinux.py
```diff
@@ -119,6 +119,30 @@ def get_config_policy(configfile):
             return line.split('=')[1].strip()
 
 
+def get_kernel_enabled(module):
+    rc, stdout, stderr = module.run_command(['grubby', '--info=ALL'])
```
Akasurde, line 123:
```suggestion
	from ansible.module_utils.common.process import get_bin_path
	...
	try:
	   grubby_bin = get_bin_path('grubby')
	except ValueError:
	   module.fail_json(msg="Failed to find the binary 'grubby' in the path")
	   
    rc, stdout, stderr = module.run_command([grubby_bin, '--info=ALL'])
```

Akasurde (changes requested):
Few comments.

WOnder93:
@Akasurde Thanks for the review, both comments should now be addressed. I solved the grubby_bin a little differently, since both the get and set methods need to use it. I also changed it to emit only a warning if the binary is not found (and skip the kernel cmdline config in that case). Please let me know if you're OK with the current code.

Akasurde:
cc @quidame @saito-hideki @justjais Could you please review this? Thanks in advance.

Akasurde (approved):
LGTM

plugins/modules/selinux.py
```diff
@@ -119,6 +120,34 @@ def get_config_policy(configfile):
             return line.split('=')[1].strip()
 
 
+def get_kernel_enabled(module, grubby_bin):
+    if grubby_bin is None:
+        module.warn("'grubby' command not found on host - kernel boot configuration won't be updated")
```
quidame, line 125: In my opinion, this breaks Ansible design at some point: kernel boot params are very important (they're able to override settings in /etc) and should not be treated with a warning. Either the module is able to change a kernel boot parameter, either it is not, and in the last case the module should fail.

It would probably be better to add a module option (boolean ?) to control this. So if, say, `update_bootloader` is *true*, and grubby is not available (because of the distribution - say, debian - or the architecture - say, arm64), the module fails.

Warning the user assumes that he conscientiously reads logs even when the playbook succeeds, whereas failing the task tells the user that the expected result is not reachable this way. Maybe I'm wrong, but that sounds more "ansible" for me.

quidame (changes requested):
@WOnder93 thanks for implementing this !

As explained right below, I would prefer to see this new feature as optional and explicit, with a dedicated module option :)

plugins/modules/selinux.py
saito-hideki: Sorry if I missed something but `grubby` makes change the behavior (for example, modifying /etc/default/grub file on RHEL8) of the selinux module beyond what the user was aware of before. So I think that we needed to mention these changes in the docs.

* https://github.com/ansible-collections/ansible.posix/blob/33e33c0af0738718170a8d85eed00d632f6490ff/plugins/modules/selinux.py#L15-L18
* https://github.com/ansible-collections/ansible.posix/blob/33e33c0af0738718170a8d85eed00d632f6490ff/plugins/modules/selinux.py#L25-L30

Also, if `selinux=0` is set by grubby, I think it is better to include this change in the result:
* https://github.com/ansible-collections/ansible.posix/blob/33e33c0af0738718170a8d85eed00d632f6490ff/plugins/modules/selinux.py#L58-L84

saito-hideki (changes requested):
@WOnder93 thank you for the PR!
I have commented about the module description and the item of the result of the module. I hope this helps!

plugins/modules/selinux.py
WOnder93: Good point - I'll add the boolean option as you are proposing.

plugins/modules/selinux.py
WOnder93: Indeed, I'm now adding a new module param for changing the kernel config, which will definitely need a documentation update.

As for including the change in the result, what exactly should be returned there? The final state after the change (disabled vs. enabled)? When no change is made to kernel config, should it not return that field or instead return some "undefined" value? The kernel config state may be inconsistent or failed to be detected, in which case it's unclear what should be returned as the final state.

saito-hideki: Hi, thank you for your clarification.
Yeah, if you add an option like `update_bootloader` and document its behavior, users understand the behavior of this option that the change will be made when the task is successful.
Also, if grubby update fails, the task will be failed with the message. So in this case I don't think it's necessary to include the changes in the return value.

plugins/modules/selinux.py
```diff
@@ -28,6 +28,11 @@
     required: true
     choices: [ disabled, enforcing, permissive ]
     type: str
+  update_kernel_param:
+    description:
+      - If set to true, will update also the kernel boot parameters when disabling/enabling SELinux.
+      - The 'grubby' tool must be present on the target system for this to work.
+    type: bool
```
saito-hideki, line 35:
```suggestion
    default: no
    type: bool
```

Hi @WOnder93!
I have checked the failed result of CI tests(https://dashboard.zuul.ansible.com/t/ansible/build/acc34e4373354f94959cf48d2b8be439).
```
ERROR: Found 1 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/selinux.py:0:0: doc-default-does-not-match-spec: Argument 'update_kernel_param' in argument_spec defines default as (False) but documentation defines default as (None)
See documentation for help: https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/validate-modules.html
```

In this case, you probably need to specify the default value. Can you try to add `default: no` in the `update_kernel_param:` section?

saito-hideki (changes requested):

plugins/modules/selinux.py
WOnder93: Of course, thanks! I forgot to check the CI results again... It sucks that I don't get email notifications when the CI finishes :/

saito-hideki (approved):
LGTM! :)

WOnder93:
@Akasurde @saito-hideki @maxamillion Is there anything else to be done before merging this? I'd be glad if i could finally tick this off as done :)

Akasurde:
> @Akasurde @saito-hideki @maxamillion Is there anything else to be done before merging this? I'd be glad if i could finally tick this off as done :)

Sure. I triggered the build.

Akasurde:
recheck

saito-hideki:
recheck

saito-hideki:
recheck

PR #161, opened 2021-03-30
fix a race condition by locking fstab with ansible's FileLock

### LINKED ISSUE #115: race condition in mount module when updating /etc/fstab in parallel
##### SUMMARY

I believe there's a race condition in the `mount` module where attempts to mount filesystems in parallel will 
clobber each other. The symptom is that only some `/etc/fstab` changes get written, which causes the
subsequent `mount` command to fail, breaking the play.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ansible.posix.mount

##### ANSIBLE VERSION

```paste below
ansible 2.9.15
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.8 (default, Apr 16 2020, 01:36:27) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]
```

##### CONFIGURATION

```paste below
DEPRECATION_WARNINGS(/etc/ansible/ansible.cfg) = False
```

##### OS / ENVIRONMENT

```
[root@kura a]# uname -a
Linux kura 4.18.0-193.28.1.el8_2.x86_64 #1 SMP Thu Oct 22 00:20:22 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
[root@kura a]# cat /etc/centos-release
CentOS Linux release 8.2.2004 (Core)
```

I'm running this on an old `Intel(R) Xeon(R) CPU E31270 @ 3.40GHz`

##### STEPS TO REPRODUCE

In this case, we have an inventory that targets the same system with multiple inventory_hostnames, but the same ansible_host.

While such a scenario might seem contrived, we've been using it very successfully as a way to configure lots of docker containers on a host. Each docker container gets a separate line in the inventory, and we drop configuration files onto the host, and bind-mount directories into each docker container during the `docker_container` task. By modifying `ansible_host` and `ansible_connection`, we can choose where the various containers should be provisioned, which is important for our use-case.
 
We hit problems though as soon as we started trying to mount different physical disks into different containers however.

Here's an example inventory file that should work anywhere:

```
[all]
foo ansible_connection=local ansible_host=localhost
bar ansible_connection=local ansible_host=localhost
baz ansible_connection=local ansible_host=localhost
ook ansible_connection=local ansible_host=localhost
moo ansible_connection=local ansible_host=localhost
dog ansible_connection=local ansible_host=localhost
cow ansible_connection=local ansible_host=localhost
cat ansible_connection=local ansible_host=localhost
pig ansible_connection=local ansible_host=localhost
eek ansible_connection=local ansible_host=localhost
see ansible_connection=local ansible_host=localhost
kit ansible_connection=local ansible_host=localhost
kat ansible_connection=local ansible_host=localhost
```

and an Ansible playbook that reproduces the problem. As this bug is a race, you might have to run it multiple times to hit the bug:

```yaml
- hosts: all
  gather_facts: false
  tasks:
  - name: create mountpoints
    file:
      state: directory
      path: /tmp/mountpoints/{{ inventory_hostname }}

  - name: mount a filesystem to our mountpoint
    mount:
      src: /etc/ansible
      state: mounted
      fstype: none
      opts: defaults,bind
      path: /tmp/mountpoints/{{ inventory_hostname }}
```

##### EXPECTED RESULTS

We should have all filesystems (in this case, just bind-mounted /etc/ansible dirs) mounted to `/tmp/mountpoint/{{ inventory_hostname }}`

##### ACTUAL RESULTS

```paste below
[root@kura a]# ansible-playbook  -i hosts test.yml

PLAY [all] ********************************************************************************************************************************************

TASK [create mountpoints] *****************************************************************************************************************************
ok: [foo]
ok: [ook]
ok: [moo]
ok: [baz]
ok: [bar]
ok: [dog]
ok: [cow]
ok: [pig]
ok: [eek]
ok: [cat]
ok: [see]
ok: [kat]
ok: [kit]

TASK [mount a filesystem to our mountpoint] *******************************************************************************************************************
fatal: [ook]: FAILED! => {"changed": false, "msg": "Error mounting /tmp/mountpoints/ook: mount: /tmp/mountpoints/ook: can't find in /etc/fstab.\n"}
changed: [moo]
changed: [foo]
changed: [bar]
changed: [baz]
changed: [cow]
changed: [dog]
fatal: [cat]: FAILED! => {"changed": false, "msg": "Error mounting /tmp/mountpoints/cat: mount: /tmp/mountpoints/cat: can't find in /etc/fstab.\n"}
changed: [pig]
changed: [eek]
changed: [kit]
changed: [see]
changed: [kat]

PLAY RECAP ********************************************************************************************************************************************
bar                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
baz                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
cat                        : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
cow                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
dog                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
eek                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
foo                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
kat                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
kit                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
moo                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
ook                        : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
pig                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
see                        : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

As a workaround, I've been able to use `lineinfile` to manually update /etc/fstab, before using the `mount` module to update the file. It looks like lineinfile has better file-locking support, which prevents us from tripping over our feet.

quidame:
##### SUMMARY

When running the `mount` module with either `state=absent`, `state=present` or `state=mounted` on several inventory hostnames addressing the same host, each instance of the module modifies system file `/etc/fstab` on the host in a way that may result in unknown/undefined/unexpected states of the file (fstab).

Fixes #115 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mount

##### ADDITIONAL INFORMATION

See the referenced issue report for rationale and usecase.

In my tests, not only the `mounted` state might cause a failure when trying to mount a mountpoint that is not recorded in fstab, but also the `absent` state may leave records referencing mountpoints (directories) removed by the module, without failure of the module itself, and leading system to fail to reboot.

Here is my test playbook:
```yaml
---
- name: test race condition with the 'mount' module
  hosts: race
  become: yes
  gather_facts: no
  tasks:
    - name: Unmount filesystem
      mount:
        path: "/tmp/mountpoints/{{ inventory_hostname.split('.')[1] }}"
        src: "/dev/mapper/vg0-test_{{ inventory_hostname.split('.')[1] }}"
        fstype: ext4
        state: absent
      retries: 1

    - name: Get results after removal
      command:
        cmd: "grep /tmp/mountpoints/{{ inventory_hostname.split('.')[1] }} /etc/fstab"
      register: grep
      changed_when: false
      failed_when: grep.rc == 0
      ignore_errors: yes

    - name: Copy file for debugging
      copy:
        src: /etc/fstab
        dest: /tmp/debug
        mode: u=rw,go=r
        owner: ksys
        group: ksys
        remote_src: yes
      run_once: yes

    - name: Restore backup
      copy:
        src: /tmp/fstab
        dest: /etc/fstab
        mode: u=rw,go=r
        owner: root
        group: root
        remote_src: yes
      run_once: yes

    - name: Mount FS over directory
      mount:
        path: "/tmp/mountpoints/{{ inventory_hostname.split('.')[1] }}"
        src: "/dev/mapper/vg0-test_{{ inventory_hostname.split('.')[1] }}"
        fstype: ext4
        state: mounted

    - name: Get results after mount
      command:
        cmd: "grep /tmp/mountpoints/{{ inventory_hostname.split('.')[1] }} /etc/fstab"
      changed_when: false
      ignore_errors: yes
```

When ran against 24 inventory hosts (that are only one VM) in parallel, 100 consecutive times for each test, with tests by:
- the main/current version of the module, varying `forks` value in ansible.cfg (1, **5**, 7, 10)
- the pull-request's version of the module, varying its lock_timeout value (0, 1, 5, 10) (this is the delay for the module to acquire the lock, after what it fails) and `forks` (**5**, 10)

it appears that:
- race increases with `forks` value (not a surprise in this case)
- setting `lock_timeout` to 0 also increases the errors rate, showing that the lock is the point
- setting `lock_timeout` to a higher value solves the issue

### DIFF
--- a/changelogs/fragments/161-fix-115_race_condition_in_mount_module.yml
+++ b/changelogs/fragments/161-fix-115_race_condition_in_mount_module.yml
@@ -1,0 +1,6 @@
+---
+bugfixes:
+  - mount - fix a race condition that might result in unknown/unexpected fstab
+    state when running the module on several inventory hostnames adressing the
+    same host. Lock it with ansible.module_utils.common.file.FileLock.lock_file()
+    (https://github.com/ansible-collections/ansible.posix/issues/115).
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -766,7 +766,7 @@ def main():
 
     lock_timeout = 2
     lock_tempdir = '/tmp'
-    lock_message = 'another instance of the module holds the %s lock' % os.path.basename(args['fstab'])
+    lock_message = 'another instance of the module holds the %s lock: timeout (%s)'
 
     _fstab = FileLock()
     try:
@@ -872,8 +872,8 @@ def main():
             else:
                 module.fail_json(msg='Unexpected position reached')
 
-    except LockTimeout:
-        module.fail_json(msg=lock_message)
+    except LockTimeout as e:
+        module.fail_json(msg=lock_message % (os.path.basename(args['fstab']), to_native(e)))
 
     # If the managed node is Solaris, convert the boot value type to Boolean
     #  to match the type of return value with the module argument.
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -193,6 +193,7 @@ from ansible_collections.ansible.posix.plugins.module_utils.mount import ismount
 from ansible.module_utils.six import iteritems
 from ansible.module_utils._text import to_bytes, to_native
 from ansible.module_utils.parsing.convert_bool import boolean
+from ansible.module_utils.common.file import FileLock, LockTimeout
 
 
 def write_fstab(module, lines, path):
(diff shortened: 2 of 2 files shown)


plugins/modules/mount.py
```diff
@@ -734,8 +735,16 @@ def main():
     name = module.params['path']
     changed = False
 
+    lock_tempdir = '/var/tmp'
```
aminvakil, line 738: Any specific reason to use `/var/tmp` instead of `/tmp`? IIRC `/var/tmp` was useful for temporary files which are needed to persists upon a reboot.

Is it necessary for these lock files to remain after a reboot?

quidame: no, it isn't. I've changed it.

quidame:
find attached the following files:
- [playbook_test_race_condition.yml.txt](https://github.com/ansible-collections/ansible.posix/files/6240200/playbook_test_race_condition.yml.txt) that produced the two others
- [test_mount_no_lock_racy_before.log.txt](https://github.com/ansible-collections/ansible.posix/files/6240206/test_mount_no_lock_racy_before.log.txt), with the current version of the module in the collection
- [test_mount_lock_no_racy_after.log.txt](https://github.com/ansible-collections/ansible.posix/files/6240201/test_mount_lock_no_racy_after.log.txt), with the patched version of the module

Each test is based on 100 consecutive runs of the playbook against a pool of 24 *samehosts*, with `forks=10` in ansible.cfg to magnify the race. The first log shows more than 300 errors (159 *failed* for errors on *mount* tasks + 153 *ignored* for inconsistencies found in fstab after *mount* succeeded), while the second one shows no error at all.

quidame:
@aminvakil could you take a look, please ? (mouse-clicking on *re-request review* button does nothing here).
@timfoster could you test this version against your usecase ?

timfoster:
Thanks for looking at this @quidame. I hacked the fix into my older Ansible installation, and it seems to be doing the right thing now - when repeatedly running/tearing down the test case from #115 on the same system I reported the bug on, we're no longer hitting the breakage.

aminvakil:
@quidame I cannot test this for a couple of days, but it's fine as far as I can see.

quidame:
@maxamillion @gundalow @Andersson007 @Akasurde @sivel @aminvakil @skvidal Hi, it could be fine to get this fixed (soon). Any advice about ansible's FileLock() implementation ?

quidame:
@jtyr could you take a look, please ?

quidame:
review_needed

quidame:
hi fellow reviewers, with the `Hide whitespace changes` checked, the diff is pretty small (`+18 -0`) :)

plugins/modules/mount.py
```diff
@@ -734,105 +735,116 @@ def main():
     name = module.params['path']
     changed = False
 
-    if state == 'absent':
-        name, changed = unset_mount(module, args)
+    lock_timeout = 2
+    lock_tempdir = '/tmp'
```
Akasurde, line 739:
```suggestion
    lock_tempdir = tempfile.mkdtemp(prefix="ansible_mount")
```

changelogs/fragments/161-fix-115_race_condition_in_mount_module.yml
```diff
@@ -0,0 +1,6 @@
+---
+bugfixes:
+  - mount - fix a race condition that might result in unknown/unexpected fstab
+    state when running the module on several inventory hostnames adressing the
```
Akasurde, line 4:
```suggestion
    state when running the module on several inventory hostnames addressing the
```

plugins/modules/mount.py
```diff
@@ -734,105 +735,116 @@ def main():
     name = module.params['path']
     changed = False
 
-    if state == 'absent':
-        name, changed = unset_mount(module, args)
+    lock_timeout = 2
```
Akasurde, line 738: Do you think we should make it configurable? Just a thought.

plugins/modules/mount.py
quidame: the point is to lock `/etc/fstab`, that can't work if each instance of the module creates a lock with a different path, making each process unaware of the locks of the others.

plugins/modules/mount.py
quidame: maybe later, if someone reports it's not enough ? In my tests, with 24 *samehosts* in parallel, there was no error reported for 100 runs of the same play, while the same tests without locking fstab report >200 errors.

plugins/modules/mount.py
quidame: an *ansible* tmp path, that we are sure it is shared by all instances of a module within a same task could be fine, but I don't know which/where...

Akasurde: I guess we can keep it `/tmp`. Later think about changing since I myself have no experience in Locking.

timfoster:
Thanks for fixing this, very much appreciated!

Andersson007:
Could someone from maintainers take a look and merge / suggest something? FYI: the module has integration tests and they have passed.

plugins/modules/mount.py
```diff
@@ -763,105 +764,116 @@ def main():
...
-
-            if res:
-                module.fail_json(msg="Error remounting %s: %s" % (name, msg))
+            else:
+                module.fail_json(msg='Unexpected position reached')
 
-        changed = True
-    else:
-        module.fail_json(msg='Unexpected position reached')
+    except LockTimeout:
```
justjais, line 875: @quidame As I haven't tested the functionality wanted to check if LockTimeout raises an exception with a msg and if yes can we not use the respective exception msg along with the custom msg.

justjais (commented):
Thanks for raising the PR and the changes LGTM, can u respond to the cosmetic comment

plugins/modules/mount.py
quidame: hi @justjais, thanks for reviewing.
I did what you suggest, this gives:
```
fatal: [test1.jeu]: FAILED! => {"changed": false, "msg": "another instance of the module holds the fstab lock: timeout (2 sec)"}
```
(in fact it was `0.5 sec` because I had to decrease the `lock_timeout` in the code to this value to reach it. I assume that when the value is `2`, it displays `2 sec` or something very close)

sivel:
As a heads up, we are planning to deprecate and remove the `FileLock` class because it has never worked properly.  When initially implemented there were many issues with it, so we reverted everywhere it was in use, but never removed the class itself.  But we do plan to remove it.

Might be good to review https://github.com/ansible/ansible/pull/52567

quidame:
arghh, bad news :( Anyway, thanks @sivel for pointing me to this. I followed almost all links and read excellent explanations about the mess of file locking on Linux. Is it planed to _remove_ the `FileLock` class, or to _replace_ it, as attempted in ansible/ansible#52567 ?

As far as I understand the `FileLock` implementation (https://github.com/ansible/ansible/blob/cf3a304ce1711ed9a19549fbf58c004acd47c06a/lib/ansible/module_utils/common/file.py#L153-L164), the lock does not apply to the file we want to protect, but to a temporary file, that is here `/tmp/ansible-fstab.lock`, not `/etc/fstab` itself. BTW, in this PR, the lock is acquired before any read access to `/etc/fstab` (https://github.com/ansible/proposals/issues/113#issuecomment-387898356).

So, if `FileLock` is removed, what to do ? I guess that implementing `flock()`, `lockf()` or `fctnl()` directly at module level is a dead end. Would we have nothing better just because nothing is perfect ? In other words, if we have to add integration tests to prove that a locking feature is working fine everywhere, why not to add the same tests before ? Sorry, it's a rhetoric question, and doesn't help a lot.

I'm just very disappointed of this situation (not especially for *this* PR or *this* module): at work, we (my team) use `throttle: 1` to avoid breakage of playbooks against _samehosts_ (may be localhost, but not only) with `user` (/etc/passwd locked), `iptables` (xtables lock), `openssh_keypair` and `known_hosts` (racy access to the same file without lock of any kind) and more. These are just bottlenecks that slow down our deployments a lot (a lock adds some negligible overhead on _samehosts_ and makes no difference on non-_samehosts_, while throttle=1 is not scalable at all and blindly applies to non-_samehosts_ as well). But maybe more important: concurrent runs of a module may fail BECAUSE OF A LOCK (due to a system command called by the module - useradd, iptables...) and concurrent runs of another module may fail (or not fail but let system or file in bad shape - that is the case of the `mount` module btw, with state=absent) BECAUSE OF NO LOCK. I can't really figure a way to solve all of them at once, and given the age of the referenced issues and PRs... how to say ? I'm not alone ?

Is there some chance to see a module implementing the lock it needs on its own ? or is there some chance to see a common class (or whatever that could be shared in any way) to reach this purpose ? or none of them ?

Unless someone has a violet magic touch to help me to convert this no-future-PR into something else, I will close it in a few days. Any other color is welcome.

PR #166, opened 2021-04-02
ansible.posix.mount: add absent_from_fstab option

dkjii-g:
##### SUMMARY
Add absent_from_fstab option to remove the entry from fstab, but not unmount or delete the folder. Ideally this would have been the behavior of absent (as to mirror the behavior of present), but for backward compatibility I added a new verbose state

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mount

##### ADDITIONAL INFORMATION
Sometimes you may not want to delete the mountpoint (e.g. if it is not currently mounted and data is in the directory, the current behavior will simply error).

### DIFF
--- a/changelogs/fragments/166_mount_absent_fstab.yml
+++ b/changelogs/fragments/166_mount_absent_fstab.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- mount - Add absent_from_fstab state
+  - mount - Add ``absent_from_fstab`` state (https://github.com/ansible-collections/ansible.posix/pull/166).
--- a/changelogs/fragments/166_mount_absent_fstab.yml
+++ b/changelogs/fragments/166_mount_absent_fstab.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- mount - Add absent_from_fstab state
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -78,9 +78,12 @@ options:
         if I(opts) is set, and the remount command fails, the module will
         error to prevent unexpected mount changes.  Try using C(mounted)
         instead to work around this issue.
+      - C(absent_from_fstab) specifies that the device mount's entry will be
+        removed from I(fstab). This option does not unmount it or delete the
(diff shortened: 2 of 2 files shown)


plugins/modules/mount.py
```diff
@@ -734,7 +737,9 @@ def main():
     name = module.params['path']
     changed = False
 
-    if state == 'absent':
+    if state == 'absent_from_fstab':
+        name, changed = unset_mount(module, args)
```
aminvakil, line 741: I'm not sure if I'm understanding this correctly, but isn't what this PR is adding for `absent_from_fstab` is the same as `absent` (calling `unset_mount` function in https://github.com/ansible-collections/ansible.posix/blob/ecd5ad53e024c1980ea358f543478df2d26df0ce/plugins/modules/mount.py#L324) ?

aminvakil (commented):
Thanks for your contribution!

Can you please also add [changelog](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)?

plugins/modules/mount.py
dkjii-g: Yes it does the same removal from fstab, but it does not unmount or remove the mountpoint after (which absent does currently).

changelogs/fragments/166_mount_absent_fstab.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- mount - Add absent_from_fstab state
```
aminvakil, line 2:
```suggestion
  - mount - Add ``absent_from_fstab`` state (https://github.com/ansible-collections/ansible.posix/pull/166).
```

plugins/modules/mount.py
aminvakil: @dkjii-g I realized what the plan is, I've got problem understanding how your code achieves so.

Akasurde:

saito-hideki:
Closing and reopening to confirm merge process

PR #182, opened 2021-04-30
Modify boot option handling on Linux systems

### LINKED ISSUE #28: [mount] Allow "boot" option on all Linux via "noauto"
##### SUMMARY
Currently `boot` option only applies to _Solaris_. It would be nice to support it on a wider range of operating systems.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mount

##### ADDITIONAL INFORMATION
- A network share might not be available at boot.
- Non-root mounting is possible when such an entry is present at `fstab`.
- It also allows to circumvent the problem that there is no module which allows mounting without editing fstab (#48134).

saito-hideki:
##### SUMMARY

Modify boot option handling on Linux systems:

* Fixes #28
* Modified behavior to set `noauto` option if `boot` is `no` on Linux system
* If `opts` contains` noauto` or `defaults`, the `boot` option will be ignored
* If `noauto` is not specified, the default behavior will be `auto`, so `boot: yes` does not explicitly set `auto`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- ansible.posix.mount

##### ADDITIONAL INFORMATION
None

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -180,6 +180,7 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.ansible.posix.plugins.module_utils.mount import ismount
 from ansible.module_utils.six import iteritems
 from ansible.module_utils._text import to_bytes, to_native
+from ansible.module_utils.parsing.convert_bool import boolean
 
 
 def write_fstab(module, lines, path):
@@ -672,7 +673,7 @@ def main():
             opts='-',
             passno='-',
             fstab=module.params['fstab'],
-            boot='yes'
+            boot='yes' if module.params['boot'] else 'no'
         )
         if args['fstab'] is None:
             args['fstab'] = '/etc/vfstab'
@@ -834,6 +835,10 @@ def main():
     else:
         module.fail_json(msg='Unexpected position reached')
 
+    # If the managed node is Solaris, convert the boot value type to Boolean
+    #  to match the type of return value with the module argument.
+    if platform.system().lower() == 'sunos':
+        args['boot'] = boolean(args['boot'])
     module.exit_json(changed=changed, **args)
 
 
--- a/changelogs/fragments/185_mount_at_boot.yml
+++ b/changelogs/fragments/185_mount_at_boot.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+- "mount - Handle ``boot`` option on Solaris correctly (https://github.com/ansible-collections/ansible.posix/issues/184)."


plugins/modules/mount.py
```diff
@@ -93,7 +93,10 @@
   boot:
     description:
       - Determines if the filesystem should be mounted on boot.
-      - Only applies to Solaris systems.
+      - Only applies to Solaris and Linux systems.
```
quidame, line 96: Hi @saito-hideki why only on Solaris and Linux ? Can't it be extended to *BSD too ? Regarding fstab format, Solaris is the only one system that differs from all others. I guess the `noauto` option is very portable, no ?

tests/integration/targets/mount/tasks/main.yml
```diff
@@ -294,3 +294,37 @@
     - /tmp/myfs.img
     - /tmp/myfs
     when: ansible_system in ('Linux')
+
+- name: Block to test boot option for Linux
+  block:
+  - name: Create empty file
+    command: dd if=/dev/zero of=/tmp/myfs.img bs=1048576 count=20
+  - name: Format FS
+    community.general.system.filesystem:
```
quidame, line 303: Is there something special with this module, that is the only one to be called by its FQCN ? BTW, `community.general.filesystem` is enough.

tests/integration/targets/mount/tasks/main.yml
```diff
@@ -294,3 +294,37 @@
     - /tmp/myfs.img
     - /tmp/myfs
     when: ansible_system in ('Linux')
+
+- name: Block to test boot option for Linux
+  block:
+  - name: Create empty file
+    command: dd if=/dev/zero of=/tmp/myfs.img bs=1048576 count=20
```
quidame, line 301: please consider using `community.general.filesize` rather than `command`:
```yaml
- name: Create empty file
  community.general.filesize:
    path: /tmp/myfs.img
    size: 20M
```

plugins/modules/mount.py
saito-hideki: Yeah that make sense, I'm planning that once this PR will be merged, I'll expand it for *BSD :)

tests/integration/targets/mount/tasks/main.yml
saito-hideki: This task uses the following task from the previous section which was written for the same purpose:

https://github.com/ansible-collections/ansible.posix/blob/main/tests/integration/targets/mount/tasks/main.yml#L232-L236

I'll modify both tasks from `community.general.system.filesystem` to `community.general.filesystem`. Thanks!

tests/integration/targets/mount/tasks/main.yml
saito-hideki: Really appreciate your advice. I'll change it from `dd` to `community.general.filesize`

saito-hideki:
@quidame I have fixed it according to your advice. I would appreciate it if you could review it. Thanks!

plugins/modules/mount.py
```diff
@@ -699,14 +714,27 @@ def main():
...
             args[key] = module.params[key]
+    if platform.system().lower() == 'linux':
+        # Linux has 'noauto' as mount opts to handle mount on boot
+        #  So boot option should manage 'noauto' in opts
+        # TODO: We need to support other system like *BSD that 'noauto' option available
+        opts = args['opts'].split(',')
+        if 'noauto' in opts:
+            args['warnings'].append("Ignore the 'boot' due to 'opts' contains 'noauto'.")
+        elif not module.params['boot']:
+            args['boot'] = 'no'
```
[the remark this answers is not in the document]
saito-hideki: Hi! thank you for the review. That was to match the return value type with Solaris. For Solaris, return value includes `boot` but it is string.
Eventually, I think that we also need to change the type of return value for Solaris. But I think it's probably better to do it in another PR.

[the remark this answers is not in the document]
saito-hideki: @quidame I have created another PR # 185. This not only resolves # 184, but also changes the `boot` parameter type to `boolean` in return values on the Solaris node.

quidame:
the changelog-fragment check seems to be fixed and should pass now

saito-hideki:

plugins/modules/mount.py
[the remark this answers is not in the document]
saito-hideki: #185 has been merged to main branch.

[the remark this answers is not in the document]
aminvakil: Could you please rebase this branch? I guess the conflicts are because of #185 you've just mentioned.

PR #187, opened 2021-05-12
synchronize: fix to honor become_user when become_method sudo

### LINKED ISSUE #186: synchronize module ignores become_user when become in effect
##### SUMMARY

The `synchronize` module always writes files as root if `become` is in effect and `become_method` is `sudo`, regardless of the specified `become_user`.  It should write files as the `become_user`.  This is quite unexpected to write the file with a different user than the one specified, and contradicts the documentation for the module:

>  The user and permissions for the synchronize `dest` are those of the `remote_user` on the destination host or the `become_user` if `become=yes` is active.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

synchronize

##### ANSIBLE VERSION

```paste below
 $ ansible --version
ansible 2.10.9
  config file = /home/scott/src/setup/.ansible.cfg
  configured module search path = ['/home/scott/src/setup/lib']
  ansible python module location = /home/scott/src/setup/venv/lib/python3.9/site-packages/ansible
  executable location = /home/scott/src/setup/venv/bin/ansible
  python version = 3.9.5 (tags/v3.9.5:0a7dcbdb13, May  4 2021, 16:51:37) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below
```

##### OS / ENVIRONMENT

recent linux on client and server

##### STEPS TO REPRODUCE

```
$ ssh scott@scott rm -f testfile
$ rm -f empty.cfg testfile
$ touch empty.cfg testfile
$ ANSIBLE_CONFIG=empty.cfg \
  env -i PATH=$PATH SSH_AUTH_SOCK=$SSH_AUTH_SOCK \
  `which ansible` \
      -b --become-user=scott \
      -i scott, \
      -m synchronize \ 
      -a "src=testfile dest=/home/scott/ archive=false" \
      scott \
   && ssh scott stat -c %U:%G testfile
 ```

##### EXPECTED RESULTS

remote file should be owned by `scott:scott` as shown by `stat` command

##### ACTUAL RESULTS

```
scott | CHANGED => {
    "changed": true,
    "cmd": "/usr/local/bin/rsync --delay-updates -F --compress --rsh=/usr/bin/ssh -S none -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null --rsync-path=sudo rsync --out-format=<<CHANGED>>%i %n%L /home/scott/src/setup/testfile scott:/home/scott/",
    "msg": "<f+++++++++ testfile\n",
    "rc": 0,
    "stdout_lines": [
        "<f+++++++++ testfile"
    ]
}

$ ssh scott@scott stat -c %U:%G testfile
root:root
```

### PR DESCRIPTION
**Pull request by smemsh**
##### SUMMARY

When `become_method` is sudo, the `synchronize` module ignores `become_user`, always running as root.  This means one cannot create files as a target user, when they need to get in via a third user and can only `sudo` via that one.  In my case, I'm connecting via a special provisioning user that has sudo privs, but I need to create the files as the `become_user`.  I'm using it to deposit skeleton files, and there should be no reason to run another task with `chown`; after all, the documentation already describes the desired behavior:

>  The user and permissions for the synchronize `dest` are those of the `remote_user` on the destination host or the `become_user` if `become=yes` is active.

This patch takes the running `become_user` (if it's not `None`) and adds it to the `sudo` command with the `-u` command line option, so the file gets created correctly.  I have tested this and it works.

Other `become_method`s are ignored, but they already were anyways (the code already has a TODO to add other methods, which we don't attempt in this patch)

Fixes #186

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

synchronize

##### ADDITIONAL INFORMATION

See reproduction in #186.

This appears to have been in place since ansible/ansible@811a906332eed12e9d3d976032341a6912b56247

### DIFF
--- a/changelogs/fragments/187-fix-synchronize-become-user.yml
+++ b/changelogs/fragments/187-fix-synchronize-become-user.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - synchronize - use become_user when invoking rsync on remote with sudo
+    (https://github.com/ansible-collections/ansible.posix/issues/186).
--- a/plugins/action/synchronize.py
+++ b/plugins/action/synchronize.py
@@ -371,7 +371,10 @@ class ActionModule(ActionBase):
                 # If no rsync_path is set, become was originally set, and dest is
                 # remote then add privilege escalation here.
                 if self._play_context.become_method == 'sudo':
-                    rsync_path = 'sudo rsync'
+                    if self._play_context.become_user:
+                        rsync_path = 'sudo -u %s rsync' % self._play_context.become_user
+                    else:
+                        rsync_path = 'sudo rsync'
                 # TODO: have to add in the rest of the become methods here
 
             # We cannot use privilege escalation on the machine running the
--- a/tests/unit/plugins/action/fixtures/synchronize/basic_become/meta.yaml
+++ b/tests/unit/plugins/action/fixtures/synchronize/basic_become/meta.yaml
@@ -25,7 +25,8 @@ asserts:
     - "self.execute_called"
     - "self.final_module_args['_local_rsync_path'] == 'rsync'"
     # this is a crucial aspect of this scenario ...
-    - "self.final_module_args['rsync_path'] == 'sudo rsync'"
+    # note: become_user None -> root
+    - "self.final_module_args['rsync_path'] == 'sudo -u root rsync'"
     - "self.final_module_args['src'] == '/tmp/deleteme'"
     - "self.final_module_args['dest'] == 'root@el6host:/tmp/deleteme'"
     - "self.task.become == True"
--- a/tests/unit/plugins/action/fixtures/synchronize/basic_become_cli/meta.yaml
+++ b/tests/unit/plugins/action/fixtures/synchronize/basic_become_cli/meta.yaml
@@ -25,7 +25,8 @@ asserts:
     - "self.execute_called"
     - "self.final_module_args['_local_rsync_path'] == 'rsync'"
     # this is a crucial aspect of this scenario ...
-    - "self.final_module_args['rsync_path'] == 'sudo rsync'"
+    # note: become_user None -> root
+    - "self.final_module_args['rsync_path'] == 'sudo -u root rsync'"
     - "self.final_module_args['src'] == '/tmp/deleteme'"
     - "self.final_module_args['dest'] == 'root@el6host:/tmp/deleteme'"
     - "self.task.become == None"

### CODE REVIEW

**Comment by smemsh** [comment]
The changelog fragment apparently has to be in the last commit? That doesn't seem right.  It means one can't add commits to the branch tip without an interactive rebase every time to reorder them.  The command it's using to determine is:
```
shell: git show --name-status --exit-code --oneline --diff-filter=A | grep changelogs/fragments/
```
but shouldn't it use instead:
```
shell: git show {{basebranch}}.. --name-status --exit-code --oneline --diff-filter=A | grep changelogs/fragments/
```

I don't know the right name for the templated branch variable ie `{{basebranch}}`, but I would think it should be using that, so it's looking at the whole diff in the patch and not just the last commit.  My changelog fragment was there and I had to reorder it to be the last commit just for the fragment detection rule to work.

File: changelogs/fragments/187-fix-synchronize-become-user.yml [hunk 1]
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
+  - synchronize - use become_user when invoking rsync on remote with sudo
+    (fixes https://github.com/ansible-collections/ansible.posix/issues/186)
```
aminvakil, line 4:
```suggestion
    (https://github.com/ansible-collections/ansible.posix/issues/186).
```

aminvakil (commented):
Thanks for your contribution!
You're right about changelog fragment check bug, I can't find the repository to open an issue/PR to fix this bug right now.

smemsh:
my patch, it was working fine, then after only a trivial change to changelog fragment was made, now this:
```
ERROR! Collection artifact at 'collections/cisco-nxos-2.2.1-dev4.tar.gz' is not a valid tar file.
```
not clear what that has to do with my patch?

aminvakil:
I would just ignore the CI :)
Feel free to close/reopen the PR to test again, as I've found some times which bot commands (rebuild_failed, recheck) does not work either.
Closing/reopening always work though ;)

smemsh:
the repository for CI bugs might be https://github.com/ansible/ansibullbot

smemsh:
recheck

smemsh:
FYI, don't see the changelog fragment detection code in ansibullbot repo anywhere, so it must be somewhere else

Akasurde:
@quidame @aminvakil Could you please review this and let me know? Thanks in advance.

aminvakil (approved):
LGTM

Akasurde:
recheck

justjais (approved):

PR #195, opened 2021-05-28
sysctl: add basic filesystem attribute setting options for sysctl_file

### LINKED ISSUE #108: sysctl custom file permissions
##### SUMMARY
 I save my custom configurations under /etc/sysctl.d/10-performance.conf (performance configs). The problem is when this file is created it has a very restrictive permission, 0600 to be exactly. Should be 0644 as the other files.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
sysctl

##### ANSIBLE VERSION

```paste below
ansible 2.10.0
  config file = /home/mhalano/projs/infra-marcos-new/ansible.cfg
  configured module search path = ['/home/mhalano/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.6 (default, Sep 25 2020, 09:36:53) [GCC 10.2.0]

```

##### CONFIGURATION

```paste below
INTERPRETER_PYTHON(/home/mhalano/projs/infra-marcos-new/ansible.cfg) = auto
```

##### OS / ENVIRONMENT

Ubuntu 20.10 but using the ansible-base package from Eoan on the official repository.

##### STEPS TO REPRODUCE
.1 Execute this minimum tasks:

```yaml
- name: Reduce Swappiness
  sysctl:
    name: vm.swappiness
    value: "10"
    reload: yes
    state: present
    sysctl_file: /etc/sysctl.d/10-performance.conf
```
##### EXPECTED RESULTS

The file /etc/sysctl.d/10-performance.conf should have as permission 0644.

##### ACTUAL RESULTS

The file /etc/sysctl.d/10-performance.conf has permission 0600 even if it is not a critical file (security key or whereveer)

```paste below
mhalano@glados:~/projs/infra-marcos-new$ ls -lh /etc/sysctl.d/10-performance.conf 
-rw------- 1 root root 42 out 30 16:34 /etc/sysctl.d/10-performance.conf
```

saito-hideki:
##### SUMMARY
sysctl: add basic filesystem attribute setting options for sysctl_file:

* Fixes #108

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- ansible.posix.sysctl

##### ADDITIONAL INFORMATION
None

### DIFF
--- a/changelogs/fragments/108_sysctl_add_mode_option.yml
+++ b/changelogs/fragments/108_sysctl_add_mode_option.yml
@@ -1,0 +1,3 @@
+minor_changes:
+- sysctl - add a basic file system attribute setting option to allow the sysctl module
+  to set the file attributes to ``sysctl_file`` (https://github.com/ansible-collections/ansible.posix/issues/108).
--- a/plugins/modules/sysctl.py
+++ b/plugins/modules/sysctl.py
@@ -56,6 +56,13 @@ options:
             - Verify token value with the sysctl command and set with -w if necessary
         type: bool
         default: 'no'
+    unsafe_writes:
+        description:
+            - This option is not used.
+
+extends_documentation_fragment:
+  - ansible.builtin.files
+
 author:
 - David CHANIAL (@davixx)
 '''
@@ -93,6 +100,22 @@ EXAMPLES = r'''
     sysctl_set: yes
     state: present
     reload: yes
+
+# Set file mode 0644 to /tmp/test_sysctl.conf
+- ansible.posix.sysctl:
+    name: net.ipv4.ip_forward
+    value: '1'
+    sysctl_file: /tmp/test_sysctl.conf
+    reload: no
+    mode: '0644'
+
+# Set file mode rw-r--r-- to /tmp/test_sysctl.conf
+- ansible.posix.sysctl:
+    name: net.ipv4.ip_forward
+    value: '1'
+    sysctl_file: /tmp/test_sysctl.conf
+    reload: no
+    mode: u=rw,g=r,o=r
 '''
 
 # ==============================================================
@@ -184,6 +207,13 @@ class SysctlModule(object):
             elif not self._values_is_equal(self.proc_value, self.args['value']):
                 self.changed = True
                 self.set_proc = True
+
+        # Set file permissions if there are differences.
+        # - Ansible 2.9.x does not support 'path' like load_file_common_arguments (params, path=PATH).
+        #   so set 'sysctl_file' as 'path' in module.params.
+        self.module.params['path'] = self.sysctl_file
+        file_args = self.module.load_file_common_arguments(self.module.params)
+        self.changed = self.module.set_fs_attributes_if_different(file_args, self.changed)
 
         # Do the work
         if not self.module.check_mode:
@@ -394,8 +424,9 @@ def main():
             reload=dict(default=True, type='bool'),
             sysctl_set=dict(default=False, type='bool'),
             ignoreerrors=dict(default=False, type='bool'),
-            sysctl_file=dict(default='/etc/sysctl.conf', type='path')
+            sysctl_file=dict(default='/etc/sysctl.conf', type='path'),
         ),
+        add_file_common_args=True,
         supports_check_mode=True,
         required_if=[('state', 'present', ['value'])],
     )
--- a/tests/integration/targets/sysctl/tasks/main.yml
+++ b/tests/integration/targets/sysctl/tasks/main.yml
@@ -216,9 +216,15 @@
 
 - name: Test on RHEL VMs
   when:
-    - ansible_facts.virtualization_type != 'docker'
-    - ansible_facts.distribution == 'RedHat'
+    - ansible_facts.virtualization_type not in ['docker', 'containerd']
+    - ansible_system == 'Linux'
   block:
+    # Initialize parameter
+    - name: Remove sysctl property using module
+      sysctl:
+        name: vm.swappiness
+        state: absent
+
     # Test reload: yes
     - name: Set sysctl property using module
       sysctl:
@@ -310,3 +316,102 @@
         that:
           - sysctl_invalid_set1 is failed
           - "'vm.mmap_rnd_bits' not in sysctl_invalid_conf_content.stdout"
+
+# Test file permissions for sysctl_file
+- name: Test to set file system permissions
+  block:
+    - set_fact:
+        output_dir_test: "{{ output_dir }}/test_sysctl"
+
+    - name: make sure our testing sub-directory does not exist
+      file:
+        path: "{{ output_dir_test }}"
+        state: absent
+
+    - name: create our testing sub-directory
+      file:
+        path: "{{ output_dir_test }}"
+        state: directory
+
+    - name: copy the example conf to the test dir
+      copy:
+        src: sysctl.conf
+        dest: "{{ output_dir_test }}/permission_test.conf"
+
+    - name: Create permission test file with 0666(check_mode)
+      sysctl:
+        name: net.ipv4.ip_forward
+        value: 1
+        mode: 0666
+        reload: no
+        sysctl_file: "{{ output_dir_test }}/permission_test.conf"
+      register: permission_test1_check_mode
+      check_mode: True
+
+    - name: Ensure permission test file has been created(check_mode)
+      assert:
+        that:
+          - permission_test1_check_mode is changed
+
+    - name: Create permission test file with 0666
+      sysctl:
+        name: net.ipv4.ip_forward
+        value: 1
+        mode: 0666
+        reload: no
+        sysctl_file: "{{ output_dir_test }}/permission_test.conf"
+      register: permission_test1
+
+    - name: Ensure permission test file has been created
+      assert:
+        that:
+          - permission_test1 is changed
+
+    - name: Get permission test file stat
+      stat:
+        path: "{{ output_dir_test }}/permission_test.conf"
+      register: stat_permission_test1
+
+    - name: Ensure file permission has been set 0666
+      assert:
+        that:
+          - stat_permission_test1.stat.mode == '0666'
+
+    - name: Modify file permission of permission test file to 0600(check_mode)
+      sysctl:
+        name: net.ipv4.ip_forward
+        value: 1
+        mode: u=rw,go=
+        reload: no
+        sysctl_file: "{{ output_dir_test }}/permission_test.conf"
+      register: permission_test2_check_mode
+      check_mode: True
+
+    - name: Ensure permission test file has been created(check_mode)
+      assert:
+        that:
+          - permission_test2_check_mode is changed
+
+    - name: Modify file permission of permission test file to 0600
+      sysctl:
+        name: net.ipv4.ip_forward
+        value: 1
+        mode: u=rw,go=
+        reload: no
+        sysctl_file: "{{ output_dir_test }}/permission_test.conf"
+      register: permission_test2
+
+    - name: Ensure permission test file has been created
+      assert:
+        that:
+          - permission_test2 is changed
+
+    - name: Get permission test file stat
+      stat:
+        path: "{{ output_dir_test }}/permission_test.conf"
+      register: stat_permission_test2
+
+    - name: Ensure file permission has been set 0600
+      assert:
+        that:
+          - stat_permission_test2.stat.mode == '0600'


plugins/modules/sysctl.py
```diff
@@ -56,6 +56,16 @@
             - Verify token value with the sysctl command and set with -w if necessary
         type: bool
         default: 'no'
+    mode:
```
quidame, line 59: why to not use the `add_file_common_args=True` and `module.load_file_common_arguments()` ? https://docs.ansible.com/ansible/latest/reference_appendices/module_utils.html

This would:
- avoid to duplicate documentation of same options
- keep consistent behavior between modules sharing such options
- extend the PR to other file attributes, ownership, selinux context...

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -310,3 +316,75 @@
...
+    - name: Ensure file permission has been set 0666
+      assert:
+        that:
+          - stat_permission_test1.stat.mode == '0666'
+
+    - name: Modify file permission of permission test file to 0600
+      sysctl:
+        name: net.ipv4.ip_forward
+        value: 1
+        mode: 'u=rw,g=,o='
```
quidame, line 372:
```suggestion
        mode: u=rw,go=
```

quidame (commented):
@saito-hideki , thanks for implementing this !

plugins/modules/sysctl.py
```diff
@@ -187,6 +214,11 @@ def process(self):
 
         # Do the work
         if not self.module.check_mode:
+            # Set file permissions if there are differences
+            if self.mode is not None:
+                if self.module.set_mode_if_different(self.sysctl_file, self.mode, False) and not self.changed:
+                    self.changed = True
```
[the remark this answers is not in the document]
saito-hideki: I have fixed it.
Thanks!

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: I have fixed it.
Thanks!!

plugins/modules/sysctl.py
saito-hideki: @quidame thank you for your valuable comments every time!
I'm currently working on fixing to use `add_file_common_args` and `load_file_common_arguments() based on your valuable advice.
For the options section in docstring, sysctl module doesn't handle all file attributes at this time, so I'm considering leaving the mode section as is.

saito-hideki:

plugins/modules/sysctl.py
```diff
@@ -46,16 +46,29 @@
               C(sysctl_file) is updated.
         type: bool
         default: 'yes'
-    sysctl_file:
+    path:
```
quidame, line 49: why this ? Is it related to this PR ?

Generally speaking, you should avoid to add aliases. I would not be surprised if other reviewers complain about this. On the other hand, I personally think that almost all modules handling files should have a `path` option to point to the file, and all other should probably have a `name` :D

Anyway, as far as I know, it is not required to name the file by `path` to make use of *file common arguments*. You can pass in argument `path=sysctl_file` to `module.load_file_common_arguments()`. In that case, the module will not return all the (plethora) stat results associated to the `path` option, all those we don't have to care about here (a lot of *user readable*, *user writable*, *user executable*, *group readable* and so on, for example). In other words, if the returned key is called `path`, we get a pretty more verbose result than if it is called `sysctl_file`.

You can take a look here for a real-life example of a target file not called `path`: https://github.com/ansible-collections/community.general/blob/549dfaae6415999755f2ca722b39e8c314c67aa1/plugins/modules/system/java_keystore.py#L263-L271

quidame (commented):
2 cents more :)

plugins/modules/sysctl.py
quidame: Another example, where the option is named `path`, but the module is aware of the size of the file in its results, so it ensures *file common arguments* are applied to (or gathered from) the actual file rather than to a symlink pointing to this file:
https://github.com/ansible-collections/community.general/blob/549dfaae6415999755f2ca722b39e8c314c67aa1/plugins/modules/files/filesize.py#L467-L477

plugins/modules/sysctl.py
```diff
@@ -187,6 +216,11 @@ def process(self):
 
         # Do the work
         if not self.module.check_mode:
+            # Set file permissions if there are differences
+            file_args = self.module.load_file_common_arguments(self.module.params)
+            if file_args['mode'] is not None and self.module.set_mode_if_different(file_args['path'], file_args['mode'], False):
```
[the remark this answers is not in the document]
saito-hideki: woooow!! thanks!!
I'll modify it! :)

tests/sanity/ignore-2.10.txt
```diff
@@ -1,3 +1,6 @@
+plugins/modules/sysctl.py validate-modules:doc-default-does-not-match-spec
+plugins/modules/sysctl.py validate-modules:parameter-type-not-in-doc
+plugins/modules/sysctl.py validate-modules:undocumented-parameter
```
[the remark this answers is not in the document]
saito-hideki: The following file_common_args. These are not used by the sysctl module. I'm looking for more reasonable way like you pointed out :)
```
ERROR: plugins/modules/sysctl.py:0:0: doc-default-does-not-match-spec: Argument 'unsafe_writes' in argument_spec defines default as (False) but documentation defines default as (None)
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'attributes' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'group' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'owner' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'selevel' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'serole' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'setype' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'seuser' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/sysctl.py:0:0: parameter-type-not-in-doc: Argument 'unsafe_writes' in argument_spec defines type as 'bool' but documentation doesn't define type
```

plugins/modules/sysctl.py
saito-hideki: I really appreciate your kind advice!
As you pointed out, this was a pretty drastic idea. So I would undo it and fix it to take advantage of the path option. :D

saito-hideki: oh, 2.9 seems not to support `path`. I'll try something.

https://docs.ansible.com/ansible/2.9/reference_appendices/module_utils.html#ansible.module_utils.basic.AnsibleModule.load_file_common_arguments

plugins/modules/sysctl.py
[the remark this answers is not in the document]
saito-hideki: done :)

plugins/modules/sysctl.py
saito-hideki: done :)

plugins/modules/sysctl.py
saito-hideki: done!

plugins/modules/sysctl.py
```diff
@@ -187,6 +213,16 @@ def process(self):
 
         # Do the work
         if not self.module.check_mode:
```
quidame, line 215: There is some design issue here. `check_mode` should show what would be done, without actually doing it, that doesn't seem to be the case here. As said by its name, `module.set_fs_attributes_if_different()` set attributes if different. It handles `check_mode` on its own and consistently (as well as `changed` status and `diff`).

plugins/modules/sysctl.py
```diff
@@ -187,6 +213,16 @@ def process(self):
...
+            # Set file permissions if there are differences.
+            # - Don't replace self.changed directly it's not only for fs attr but also sysctl param changes.
+            # - Ansible 2.9.x does not support 'path' like load_file_common_arguments (params, path=PATH).
+            #   so set 'sysctl_file' as 'path' in module.params.
+            self.module.params['path'] = self.sysctl_file
+            file_args = self.module.load_file_common_arguments(self.module.params)
+            fs_attr_changed = self.module.set_fs_attributes_if_different(file_args, False)
+
+            if fs_attr_changed:
+                self.changed = True
```
quidame, line 225: This is the way it works
```suggestion
            self.changed = self.module.set_fs_attributes_if_different(file_args, self.changed)

```

quidame (commented):
thanks for improving it !

plugins/modules/sysctl.py
saito-hideki: Thank you for your detailed instruction! I have fixed it.
And I'm going to re-read the code and document links that you gave me before. thanks again!

tests/sanity/ignore-2.10.txt
quidame, line 3:
```suggestion
```

tests/sanity/ignore-2.11.txt
```diff
@@ -1,3 +1,6 @@
+plugins/modules/sysctl.py validate-modules:doc-default-does-not-match-spec
+plugins/modules/sysctl.py validate-modules:parameter-type-not-in-doc
+plugins/modules/sysctl.py validate-modules:undocumented-parameter
```
quidame, line 3:
```suggestion
```

tests/sanity/ignore-2.12.txt
```diff
@@ -1,3 +1,6 @@
+plugins/modules/sysctl.py validate-modules:doc-default-does-not-match-spec
+plugins/modules/sysctl.py validate-modules:parameter-type-not-in-doc
+plugins/modules/sysctl.py validate-modules:undocumented-parameter
```
quidame, line 3:
```suggestion
```

quidame (changes requested):
all these parameters should be handled smartly by the `add_file_common_args=True` in module's `argument_spec`. I guess the reported errors come from a draft version of this PR:
```
ERROR: plugins/modules/sysctl.py:0:0: doc-default-does-not-match-spec: Argument 'unsafe_writes' in argument_spec defines default as (False) but documentation defines default as (None)
```
since `unsafe_writes` is neither in `argument_spec` nor in `DOCUMENTATION`, this error message is not related to the current content of the PR. Please remove all tests/sanity overrides that have been added.

tests/sanity/ignore-2.10.txt
saito-hideki: Removed three lines.

tests/sanity/ignore-2.11.txt
saito-hideki: Removed three lines.

tests/sanity/ignore-2.12.txt
saito-hideki: Removed three lines.

saito-hideki:

saito-hideki:
Hi @quidame, thank you for your help :)
I have removed three lines from tests/sanity/ignore-2.1[012].txt . After that, sanity tests still failed like below [1]:

[1] https://dashboard.zuul.ansible.com/t/ansible/build/f2f338fd5aae40959725cb8dce866bce

Sorry if I missed something, But looking at the failure of the above sanity test, In order to solve this, it seems that the options of [2] need to be included in DOCUMENTS as described in [3]. I thought that `extends_documentation_fragment: files`, like the template module[4] probably solve this, but in this case,  there is another problem that the `ansible-doc ansible.posix.sysctl` command will display some unused options like `unsafe_writes`.  

[2] https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/doc_fragments/files.py
[3] https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#adding-file-options
[4] https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/template.py#L37-L41

Is there a way to solve this problem with `add_file_common_args=True`? I would appreciate it if you could give me some advice :)

Thanks!

quidame:
> Is there a way to solve this problem with `add_file_common_args=True`? I would appreciate it if you could give me some advice :)

Yes, there is :)
1. Add this in `DOCUMENTATION`:
```yaml
extends_documentation_fragment:
  - ansible.builtin.files
```
2. Overwrite some option description, type or default value as needed
```yaml
options:
  unsafe_writes:
    description:
      - This option is not used.
```
Also note that you could remove the `mode` option from `DOCUMENTATION`: it overwrites the one from file common arguments, and maybe it is not needed ?

saito-hideki:
Closing and re-opening for CI trigger.Closing and re-opening for CI trigger.

saito-hideki:
@quidame coool! thank you for your advice! :)
I have added options and remove `mode`. Now sanity test succeeded without errors!

plugins/modules/sysctl.py
```diff
@@ -56,6 +56,34 @@
...
+            - This option is not used.
+    serole:
+        description:
+            - This option is not used.
+    setype:
+        description:
+            - This option is not used.
+    seuser:
+        description:
+            - This option is not used.
```
quidame, line 79: Unlike `unsafe_writes`, which is about the *process* of modifying a file, all file properties (`mode`, `group`, `owner`, `attributes`...) are supported by the call to `module.set_fs_attributes_if_different()`, which is about file's *static* metadata. So, all these options are fully supported, as well as `mode`, and should not be overridden this way here.
```suggestion
```

plugins/modules/sysctl.py
```diff
@@ -187,6 +231,13 @@ def process(self):
 
         # Do the work
         if not self.module.check_mode:
+            # Set file permissions if there are differences.
+            # - Ansible 2.9.x does not support 'path' like load_file_common_arguments (params, path=PATH).
+            #   so set 'sysctl_file' as 'path' in module.params.
+            self.module.params['path'] = self.sysctl_file
+            file_args = self.module.load_file_common_arguments(self.module.params)
+            self.changed = self.module.set_fs_attributes_if_different(file_args, self.changed)
```
quidame, line 239: `module.set_fs_attributes_if_different()` supports `check_mode` on its own. This call should take place outside of the `if not self.module.check_mode:` branch, so when playing the module in `check_mode`, if there are differences to apply in file properties, the `set_fs_attributes_if_different()` is aware of them and will report the change, and will report the *same* change, with `check_mode=true` or `false`.
This is currently not the case: with the call to `set_fs_attributes_if_different()` skipped if `check_mode=true`, the module is unable to compute/return changes in file `mode` or `owner`.

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -310,3 +316,75 @@
...
+        value: 1
+        mode: 0666
+        reload: no
+        sysctl_file: "{{ output_dir_test }}/permission_test.conf"
+      register: permission_test1
+
+    - name: Ensure permission test file has been created
+      assert:
+        that:
+          - permission_test1.changed
```
quidame, line 356:
```suggestion
          - permission_test1 is changed
```

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -310,3 +316,75 @@
...
+        value: 1
+        mode: u=rw,go=
+        reload: no
+        sysctl_file: "{{ output_dir_test }}/permission_test.conf"
+      register: permission_test2
+
+    - name: Ensure permission test file has been created
+      assert:
+        that:
+          - permission_test2.changed
```
quidame, line 380:
```suggestion
          - permission_test2 is changed
```

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -310,3 +316,75 @@
         that:
           - sysctl_invalid_set1 is failed
           - "'vm.mmap_rnd_bits' not in sysctl_invalid_conf_content.stdout"
+
+# Test file permissions for sysctl_file
+- name: Test to set file system permissions
+  when:
+    - ansible_facts.virtualization_type != 'docker'
+    - ansible_facts.distribution == 'RedHat'
```
quidame, line 324: why just RedHat ? Is it not supposed to work on any Linux distribution ?
```suggestion
    - ansible_facts.system == 'Linux'
```

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -310,3 +316,75 @@
...
+      file:
+        path: "{{ output_dir_test }}"
+        state: directory
+
+    - name: copy the example conf to the test dir
+      copy:
+        src: sysctl.conf
+        dest: "{{ output_dir_test }}/permission_test.conf"
+
+    - name: Create permission test file wirh 0666
```
quidame, line 344:
```suggestion
    - name: Create permission test file with 0666
```

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -310,3 +316,75 @@
...
+      stat:
+        path: "{{ output_dir_test }}/permission_test.conf"
+      register: stat_permission_test1
+
+    - name: Ensure file permission has been set 0666
+      assert:
+        that:
+          - stat_permission_test1.stat.mode == '0666'
+
+    - name: Modify file permission of permission test file to 0600
```
quidame, line 368: could you please also include the same task with `check_mode=true`, and check that the results are consistent ?

quidame (changes requested):
there is still an issue with `check_mode`, when applied to new *file common parameters*, as `mode`

plugins/modules/sysctl.py
saito-hideki: > module.set_fs_attributes_if_different() supports check_mode on its own.

Thank you for your advice! I have understood that set_fs_attributes_if_different already have check logic it self using `check_file_absent_if_check_mode()`

* https://github.com/ansible/ansible/blob/96d2f79e4c349a7cbff4e5aaf11c39bac75a474e/lib/ansible/module_utils/basic.py#L1177-L1178

So I'll move this logic outside of `if not self.module.check_mode:` block.

plugins/modules/sysctl.py
saito-hideki: done! :)

plugins/modules/sysctl.py
saito-hideki: done :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: done :) :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: done! :) :) :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: sure! will do! :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: done :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: done :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: sure! will do!

quidame:
@saito-hideki I think CI failures are not really related to this PR. It seems there is some issue somewhere else...

In the failed tests on Docker, all tasks in the block with the condition
```yaml
  when:                           
    - ansible_facts.virtualization_type == 'docker' or ansible_facts.virtualization_type == 'container'
```
are skipped, whereas tasks in the block with the condition
```yaml
  when:
    - ansible_facts.virtualization_type != 'docker'
    - ansible_facts.system == 'Linux'
```
are played...

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: done!

changelogs/fragments/108_sysctl_add_mode_option.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+- sysctl - add a basic file system attribute setting option to allow the sysctl module
+  to set the file attributes to `` sysctl_file`` (https://github.com/ansible-collections/ansible.posix/issues/108).
```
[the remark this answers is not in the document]
saito-hideki: thank you for the review. I have fixed it :)

saito-hideki:
@quidame I really appreciate your valuable advice!
I have removed the following `when` section from the testing block(name: Test to set file system permissions):
```YAML
  when:
    - ansible_facts.virtualization_type != 'docker'
    - ansible_facts.system == 'Linux'
```

Also, added `needs/privileged` to `tests/integration/targets/sysctl/aliases` file to avoid "Read-only file system" error during integration tests on opensuse15. Now CI tests succeeded without errors.

Thanks again! :)

quidame:
> Also, added needs/privileged to tests/integration/targets/sysctl/aliases file to avoid "Read-only file system" error during integration tests on opensuse15. Now CI tests succeeded without errors.

Now CI tests succeeded without error because they are entirely skipped:
```
WARNING: Excluding tests marked "needs/privileged" which require --docker-privileged to run under docker: sysctl
```
It makes that the first part of the tests (which clearly target docker) are never played. I also makes that `sysctl` module is never tested on CentOS, OpenSUSE, Fedora or Ubuntu.

Sorry, I didn't realized that for this collection, Linux remote targets are *always* RedHat, and Linux docker targets are *never* RedHat, and that's probably why the condition was initially:
```yaml
  when:
    - ansible_facts.virtualization_type != 'docker'
    - ansible_facts.distribution == 'RedHat'
```
So, now I think it would be better to keep this condition (and remove the `needs/privileged` in `aliases`) and add some valuable comment to explain to next contributors why this condition shouldn't be modified. My apologies for the rework.

There is some issue with `ansible_facts.virtualization_type == 'docker'` on *some* hosts (for the record: CentOS 8 and OpenSUSE 15py3 on Docker with Ansible devel or 2.11), and this issue is not related to this PR. This issue makes that on a few docker hosts, all tests are skipped, but I would prefer that rather than a more global condition that makes that on all docker hosts, all tests are skipped.

Thanks for your patience.

tests/integration/targets/sysctl/aliases
```diff
@@ -2,3 +2,4 @@ shippable/posix/group1
 skip/aix
 skip/freebsd
 skip/osx
+needs/privileged
```
quidame, line 5:
```suggestion
```

tests/integration/targets/sysctl/tasks/main.yml
```diff
@@ -217,8 +217,14 @@
 - name: Test on RHEL VMs
   when:
     - ansible_facts.virtualization_type != 'docker'
-    - ansible_facts.distribution == 'RedHat'
+    - ansible_facts.system == 'Linux'
```
quidame, line 220: please try with:
```yaml
    when:
      - ansible_facts.virtualization_type not in ['docker', 'containerd']
```
and no more (other hosts than Linux are skipped from `aliases`)

quidame (changes requested):
@saito-hideki I've investigated about unexpected CI failures related to `ansible_facts.virtualization_type`, which seemed to not be `docker` on all docker hosts. It appears that it is `docker` or `containerd` (see https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=22937&view=logs&j=bec670ca-7c5a-5b61-508e-aa38d51599af&t=428552c5-23f1-5155-8079-dbf71dab0b75&l=277)
Could you please adjust the related `when` conditions accordingly ? (also line 25)

tests/integration/targets/sysctl/aliases
saito-hideki: yup, thanks!
I have removed it :)

tests/integration/targets/sysctl/tasks/main.yml
saito-hideki: @quidame 
I really appreciate your help! It seems that OSX doesn't support `systemctl -p` option. However, I think the `-p` option is a requirement of the sysctl module.
So I think we probably need to skip OSX as well. I would appreciate it if you could give me your opinion :)
* https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=22948&view=logs&j=a672a2ca-850f-5934-0d20-eacaaa0e01bc&t=79862c18-8b9e-5d82-2fb6-545ffdaaf722&l=867

quidame: oh yes, my first hypothesis is that (for whatever reason) `skip/macos` is needed in addition to `skip/osx` in `tests/integration/targets/sysctl/aliases` to skip **all** Mac OS X on the CI platform :)

If it doesn't work with this, so you could add a condition such as `ansible_system == 'Linux'` in the related blocks/tasks.

Or even do both, as you feel. This is not so much important at this step. We just know that we target Linux: saying it in `aliases` is less resources-consuming than saying it in the tasks.

saito-hideki: yup, thanks a lot!
I have added `ansible_system == 'Linux'` :)

Issue #175, opened 2021-04-15
Confusing doc on synchronize delete option


rdd:
I found the following doc line confusing:
https://github.com/ansible-collections/ansible.posix/blob/707523a7637ef23df21c11a6bffd90c7d0c7878e/plugins/modules/synchronize.py#L79

My rsync manpage tells me that the `--delete-excluded` flag "also delete excluded files from dest dirs", so quite the opposite of ignoring excluded files.
Should it say `--delete-after` instead which seems to be the rsync flag used when the `delete` option is used?

Akasurde:
@rdd Thanks for reporting this issue. Could you please check if #201 works for you and let me know? Thanks.

needs_info

PR #201, opened 2021-06-04
synchronize: fix misc typo

Akasurde:
##### SUMMARY

Docs should read ``--delete-after`` instead of ``--delete-excluded``.

Fixes: #175

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/synchronize.py

### DIFF
--- a/changelogs/fragments/175_synchronize.yml
+++ b/changelogs/fragments/175_synchronize.yml
@@ -1,0 +1,3 @@
+---
+trivial:
+- synchronize - fix typo in ``delete`` parameter (https://github.com/ansible-collections/ansible.posix/issues/175).
--- a/plugins/modules/synchronize.py
+++ b/plugins/modules/synchronize.py
@@ -75,7 +75,7 @@ options:
     description:
       - Delete files in C(dest) that don't exist (after transfer, not before) in the C(src) path.
       - This option requires C(recursive=yes).
-      - This option ignores excluded files and behaves like the rsync opt --delete-excluded.
+      - This option ignores excluded files and behaves like the rsync opt --delete-after.
     type: bool
     default: no
   dirs:
--- a/plugins/modules/synchronize.py
+++ b/plugins/modules/synchronize.py
@@ -73,9 +73,9 @@ options:
     default: no
   delete:
     description:
-      - Delete files in C(dest) that don't exist (after transfer, not before) in the C(src) path.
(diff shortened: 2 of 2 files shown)


changelogs/fragments/175_synchronize.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
aminvakil, line 2:
```suggestion
trivial:
```

aminvakil (commented):
I'm not sure if docs changes needs a changelog fragment at all, but if needs for the zuul bug which needs a changelog fragment, maybe change this to `trivial`?

Disclaimer: I've just recently learned `trivial` from @felixfontein from another PR and do not understood completely its use cases, and it's not been mentioned in https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment.

aminvakil (approved):
LGTM

felixfontein:
@aminvakil check out the devel docs, not the latest docs: https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs (the backport https://github.com/ansible/ansible/pull/74753 has been merged, but it will only be visible on the next deploy, which is probably next week after the Ansible 4.1.0 release).

plugins/modules/synchronize.py
```diff
@@ -75,7 +75,7 @@
     description:
       - Delete files in C(dest) that don't exist (after transfer, not before) in the C(src) path.
       - This option requires C(recursive=yes).
-      - This option ignores excluded files and behaves like the rsync opt --delete-excluded.
+      - This option ignores excluded files and behaves like the rsync opt --delete-after.
```
felixfontein, line 78:
```suggestion
      - This option ignores excluded files and behaves like the rsync opt C(--delete-after).
```

plugins/modules/synchronize.py
```diff
@@ -75,7 +75,7 @@
     description:
       - Delete files in C(dest) that don't exist (after transfer, not before) in the C(src) path.
       - This option requires C(recursive=yes).
```
felixfontein, line 77: And while I'm at it:
```suggestion
      - Delete files in I(dest) that don't exist (after transfer, not before) in the I(src) path.
      - This option requires I(recursive=yes).
```

felixfontein (approved):

PR #196, opened 2021-05-29
Modify boot option handling on Linux systems

saito-hideki:
##### SUMMARY

Rebased and recreate pull request based on #182.

* Fixes #28
* Modified behavior to set noauto option if boot is 'no' on Linux system
* Modified integration test to use filesize module instead of dd

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- ansible.posix.mount

##### ADDITIONAL INFORMATION
I created this new PR because I accidentally closed #182.

PR #203, opened 2021-06-08
Modify boot option handling on BSD systems

### LINKED ISSUE #28: [mount] Allow "boot" option on all Linux via "noauto"

saito-hideki:
##### SUMMARY

Modify boot option handling on FreeBSD, NetBSD and OpenBSD systems

* Fixes #28 for BSD systems
* Porting PR #196 to BSD systems

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- plugins/modules/mount.py

##### ADDITIONAL INFORMATION
Tested on the following platforms:
- FreeBSD 13.0
- NetBSD 9.1
- OpenBSD 6.9

### DIFF
--- a/changelogs/fragments/203_boot_opt_for_bsd.yml
+++ b/changelogs/fragments/203_boot_opt_for_bsd.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - mount - Change behavior of ``boot`` option to set ``noauto`` on BSD nodes
+    (https://github.com/ansible-collections/ansible.posix/issues/28).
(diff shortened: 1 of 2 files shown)


Akasurde:
@saito-hideki Thanks for the contribution.

Issue #178, opened 2021-04-20
Create missing mountpoint without permissions?


c0xc:
##### SUMMARY

In some environments, it's common practice to create mountpoints without read/write permissions to prevent admins and services from using them as local directories, filling up the local disk. With Ansible's mount module, a missing mountpoint is created with the default umask instead, so it may be mistaken for a mounted share if the mount failed. Is it possible to have missing mountpoints created without permissions?

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

mount

##### ADDITIONAL INFORMATION

Manual approach:
```
mkdir /mnt/xyz && chmod 0 /mnt/xyz
mount /mnt/xyz
```

Ansible workaround:
```
- name: Check for /mnt/xyz
  stat:
    path: /mnt/xyz
  register: st_mount_xyz
  ignore_errors: "yes"

- name: Create /mnt/xyz
  file:
    path: /mnt/xyz
    state: directory
    owner: root
    group: root
    mode: 0000
  when:
    - st_mount_xyz.stat is defined
    - st_mount_xyz.stat.exists == false

- name: Mount /mnt/xyz
  mount:
    src: "filer:/data/xyz"
    path: /mnt/xyz
    state: mounted
    fstype: nfs

```

https://github.com/ansible/ansible/issues/74345

quidame:
@c0xc, thanks for reporting this. For me, it's also related to #163 since these two issues refer to permissions/ownership of directories created by the module in a *one-shot* way, with no user control over what is done. Especially in this case, once the filesystem is mounted, the underlying directory (same path than the mountpoint, and hidden by the mount) is nearly inaccessible, or at least non trivial to modify. Implementing control/changes over this directory mode at playbook-level or role-level is no less than tricky, so it seems fair to implement it in the module.

PR #209, opened 2021-06-12
Add umask option for mount module

### LINKED ISSUE #163: umask option for the ansible.posix.mount module
##### SUMMARY
When using the ansible.posix.mount module to mount a filesystem under a directory structure which is not existent, the module creates the directories with the umask which is defined for the root user.
If the default umask for root is set that "other" has no permissions on newly created directories, the mount module mounts the filesystem - but no one can access it other then root because traversing the directories is not possible.

for example you want a filesystem mounted under "/myapps/app1" and both directories are non-existent:
```
[root@server ~]# ls -ld /myapps{,/app1}
ls: cannot access '/myapps': No such file or directory
ls: cannot access '/myapps/app1': No such file or directory
[root@server ~]#
```

umask is set to `0027`
```
[root@server ~]# umask; umask -S
0027
u=rwx,g=rx,o=
[root@server ~]#
```

Using ansible to mount the filesystem
```
[root@server ~]# ansible localhost -m mount -a "src=tmpfs path=/myapps/app1 fstype=tmpfs state=mounted"
localhost | CHANGED => {
    "changed": true,
    "dump": "0",
    "fstab": "/etc/fstab",
    "fstype": "tmpfs",
    "name": "/myapps/app1",
    "opts": "defaults",
    "passno": "0",
    "src": "tmpfs"
}
[root@server ~]#
```

Now the module created the directory  "/myapps/" and "/myapps/app1/" and the Filesystem is mounted under `/myapps/app1`

```
[root@server ~]# ls -ld /myapps{,/app1}
drwxr-x---. 3 root root 18 Apr  1 09:24 /myapps
drwxrwxrwt. 2 root root 40 Apr  1 09:24 /myapps/app1
[root@server ~]#
```

But the Directory /myapps/ got the permissions *750* and I can't access the mounted filesystem as a user != root - no matter what the mounted filesystem has for permissions

```
[user@server ~]$ ls -l /myapps{,/app1}
ls: cannot access '/myapps/app1': Permission denied
ls: cannot open directory '/myapps': Permission denied
[user@server ~]$
```

An option to avoid this situation would be to specify the umask on calling the module like with the "git" module.
if you could set umask=0022 - all directories which gets created automatically would be created that you can traverse these directories.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
ansible.posix.mount

##### ADDITIONAL INFORMATION

```yaml
 ansible localhost -m mount -a "umask=0022 src=tmpfs path=/myapps/app1 fstype=tmpfs state=mounted"
```
That would cause ansible to set the umask for that specific task and the parent directories would get the propper permission - if this is wanted.

satken2:
##### SUMMARY

I added `mode` option for mount module.
When this module creates new directories for mount point, this value is used as permission of the directory.

Fixes #163, #178

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- mount

##### ADDITIONAL INFORMATION

When we want to set specific permission to the mount point directory, we can use `file` module. 
However once the filesystem is mounted, the underlying directory is nearly inaccessible, and cannot be modified by `file` module.
This is a bit inconvenient, so I added `mode` option so that permission can be set before actual mount operation.

In #163,  author requests to add `umask` option but when the mount point does not exist, this module creates directories, not files. So I think `mode` is more appropriate than `umask`.

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -105,6 +105,17 @@ options:
         the original file back if you somehow clobbered it incorrectly.
     type: bool
     default: no
+  mode:
+    description:
+      - The permission applied to create a new directory for mount point.
+        If the mount point already exists, this parameter is not used.
+      - This parameter only affects to the mount point itself.
+        If this module creates multiple directories recursively,
+        other directories follow the system's default umask.
+      - Note that after running this task and device being successfuly mounted,
+        the mode of the original directory will be hidden by the target device.
+    type: raw
+    required: false
 notes:
   - As of Ansible 2.3, the I(name) option has been changed to I(path) as
     default, but I(name) still works as well.
@@ -122,6 +133,7 @@ EXAMPLES = r'''
     fstype: iso9660
     opts: ro,noauto
     state: present
+    mode: 0755
 
 - name: Mount up device by label
   ansible.posix.mount:
@@ -665,6 +677,7 @@ def main():
             src=dict(type='path'),
             backup=dict(type='bool', default=False),
             state=dict(type='str', required=True, choices=['absent', 'mounted', 'present', 'unmounted', 'remounted']),
+            mode=dict(type='raw'),
         ),
         supports_check_mode=True,
         required_if=(
@@ -761,6 +774,7 @@ def main():
 
     state = module.params['state']
     name = module.params['path']
+    mode = module.params['mode']
     changed = False
 
     if state == 'absent':
@@ -812,6 +826,9 @@ def main():
                             # check above. As long as it's a dir, we don't need to error out.
                             if not (ex.errno == errno.EEXIST and os.path.isdir(b_curpath)):
                                 raise
+
+                if mode is not None:
+                    os.chmod(name, int(mode))
 
             except (OSError, IOError) as e:
                 module.fail_json(
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -827,12 +827,17 @@ def main():
                             if not (ex.errno == errno.EEXIST and os.path.isdir(b_curpath)):
                                 raise
 
-                if mode is not None:
-                    os.chmod(name, int(mode))
-
             except (OSError, IOError) as e:
                 module.fail_json(
                     msg="Error making dir %s: %s" % (name, to_native(e)))
+
+            # Set permissions to the newly created mount point.
+            if mode is not None:
+                try:
+                    changed = module.set_mode_if_different(name, mode, changed)
+                except Exception as e:
+                    module.fail_json(
+                        msg="Error setting permissions %s: %s" % (name, to_native(e)))
 
         name, backup_lines, changed = _set_mount_save_old(module, args)
         res = 0
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -115,7 +115,7 @@ options:
       - Note that after running this task and device being successfuly mounted,
         the mode of the original directory will be hidden by the target device.
     type: raw
-    required: false
+    version_added: '1.3.0'
 notes:
   - As of Ansible 2.3, the I(name) option has been changed to I(path) as
     default, but I(name) still works as well.
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -105,13 +105,10 @@ options:
         the original file back if you somehow clobbered it incorrectly.
     type: bool
     default: no
-  mode:
+  umask:
     description:
-      - The permission applied to create a new directory for the mount point.
+      - The permission applied to create new directory(ies) for the mount point.
         If the mount point already exists, this parameter is not used.
-      - This parameter only affects the mount point itself.
-        If this module creates multiple directories recursively,
-        other directories follow the system's default umask.
       - Note that after running this task and the device being successfully mounted,
         the mode of the original directory will be hidden by the target device.
     type: raw
@@ -133,7 +130,7 @@ EXAMPLES = r'''
     fstype: iso9660
     opts: ro,noauto
     state: present
-    mode: 0755
+    umask: 0022
 
 - name: Mount up device by label
   ansible.posix.mount:
@@ -677,7 +674,7 @@ def main():
             src=dict(type='path'),
             backup=dict(type='bool', default=False),
             state=dict(type='str', required=True, choices=['absent', 'mounted', 'present', 'unmounted', 'remounted']),
-            mode=dict(type='raw'),
+            umask=dict(type='raw'),
         ),
         supports_check_mode=True,
         required_if=(
@@ -774,7 +771,7 @@ def main():
 
     state = module.params['state']
     name = module.params['path']
-    mode = module.params['mode']
+    umask = module.params['umask']
     changed = False
 
     if state == 'absent':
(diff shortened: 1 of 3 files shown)


plugins/modules/mount.py
```diff
@@ -813,6 +827,9 @@ def main():
                             if not (ex.errno == errno.EEXIST and os.path.isdir(b_curpath)):
                                 raise
 
+                if mode is not None:
+                    os.chmod(name, int(mode))
```
Akasurde, line 831: I have couple of comments over here -

- Move this block out try-except to its own try-except. So that we don't mask the original exception.
- Please use [set_mode_if_different](https://github.com/ansible/ansible/blob/cf3a304ce1711ed9a19549fbf58c004acd47c06a/lib/ansible/module_utils/basic.py#L885) instead of `chmod`.
- Other modules provides `mode` such as `u+rw,g-wx,o-rwx` or `u=rw,g=r,o=r`, please add this support as well.

Akasurde (changes requested):

Akasurde:
cc @quidame @aminvakil @justjais

plugins/modules/mount.py
satken2: Thank you for your advise.
I fixed my code based on the comments.

plugins/modules/mount.py
```diff
@@ -105,6 +105,17 @@
...
+    description:
+      - The permission applied to create a new directory for mount point.
+        If the mount point already exists, this parameter is not used.
+      - This parameter only affects to the mount point itself.
+        If this module creates multiple directories recursively,
+        other directories follow the system's default umask.
+      - Note that after running this task and device being successfuly mounted,
+        the mode of the original directory will be hidden by the target device.
+    type: raw
+    required: false
```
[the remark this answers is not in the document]
satken2: Thank you for suggestion. I fixed my code according to the suggestion, but I replaced tabs by spaces.

plugins/modules/mount.py
```diff
@@ -817,6 +831,14 @@ def main():
                 module.fail_json(
                     msg="Error making dir %s: %s" % (name, to_native(e)))
 
+            # Set permissions to the newly created mount point.
+            if mode is not None:
+                try:
+                    changed = module.set_mode_if_different(name, mode, changed)
```
aminvakil, line 837: This does not work on non-Linux operating systems? I've seen you run integration tests only on Linux systems when it was using `chmod`, I don't know about this function.
If so, please also mention this on parameter notes, and maybe warn user when the target os is not Linux?

tests/integration/targets/mount/tasks/main.yml
```diff
@@ -332,3 +332,44 @@
     - /tmp/myfs.img
     - /tmp/myfs
   when: ansible_system in ('Linux')
+
```
aminvakil, line 335: Please add a integration test which uses something like `u+rw,g-wx,o-rwx` or `u=rw,g=r,o=r`.

aminvakil (commented):
Thanks for adding this feature!

quidame:
Hi @satken2, thanks for your contribution !

IMHO, this PR doesn't fix #163, as said in its description: the new parameter only applies to the mountpoint itself, not its parents, even if they're created by the module, that means that if `/srv` is empty and the user wants to mount some device at `/srv/foo/bar/public` with a public access to it (say `mode: u=rwx,go=rx`), this mountpoint will still be unreachable by most users because `/srv/foo` and `/srv/foo/bar` have get mode `0700`.

quidame:
> ##### SUMMARY
> 
> I added `mode` option for mount module.
> When this module creates new directories for mount point, this value is used as permission of the directory.
> 
> Fixes #163, #178
> ##### ISSUE TYPE
> 
>     * Feature Pull Request
> 
> 
> ##### COMPONENT NAME
> 
>     * mount
> 
> 
> ##### ADDITIONAL INFORMATION
> 
> When we want to set specific permission to the mount point directory, we can use `file` module.
> However once the filesystem is mounted, the underlying directory is nearly inaccessible, and cannot be modified by `file` module.
> This is a bit inconvenient, so I added `mode` option so that permission can be set before actual mount operation.
> 
> In #163, author requests to add `umask` option but when the mount point does not exist, this module creates directories, not files. So I think `mode` is more appropriate than `umask`.

I think there is a confusion here. `umask` applies to both files and directories. To directories, with the exact value of the mask; to files, with the value of the mask and by zeroing all executable bits.

satken2:
@quidame Thank you for your advice. Certainly, when creating directories recursively, I have to set permissions on all directories to resolve #163, as you point out. So I will fix my code.

>I think there is a confusion here. umask applies to both files and directories. To directories, with the exact value of the mask; to files, with the value of the mask and by zeroing all executable bits.

The `mount` module only creates directories, so I don't think it is necessary to consider creating files, I guess.
So I added `mode` option.

tests/integration/targets/mount/tasks/main.yml
```diff
@@ -332,3 +332,44 @@
...
+      that:
+      - mount_point_stat['stat']['mode'] == '0000'
+  - name: Remove the test FS
+    file:
+      path: '{{ item }}'
+      state: absent
+    loop:
+    - /tmp/myfs.img
+    - /tmp/myfs
+  when: ansible_system in ('Linux')
```
quidame, line 375: why this restriction ? Does the new `mode` parameter not support other OSes ?

quidame:
> The `mount` module only creates directories, so I don't think it is necessary to consider creating files, I guess.
> So I added `mode` option.

There is no need to create files to use umask. It applies to almost any *file* (regular files, directories, unix sockets, named pipes or fifos, character devices...) except symlinks, at creation time. Even if at the end it doesn't change anything to your code, please consider `umask` as a plausible option. It seems that you have discarded it on a misunderstanding.

For me, `mode` sounds more like something I (user) would have control over, at any time, not only at creation time. (ok, in this case, not at any time, but every time the underlying directory used as mountpoint is not mounted, and exposed/unhidden). At least 2 modules make use of `umask` (https://github.com/ansible/ansible/blob/ec408a69f19052190f1766a357c882fc86cfeada/lib/ansible/modules/git.py#L136-L141, https://github.com/ansible/ansible/blob/ec408a69f19052190f1766a357c882fc86cfeada/lib/ansible/modules/pip.py#L106-L114). I don't know if there are modules using `mode` in a one-shot way, that is the purpose of `umask`, and if there are, I would probably consider them as *confusing*, and still suggest: please consider `umask` for this new option. If a contributor needs `mode` to implement mode of the root of the mounted filesystem, and `mode` is already used for something else that behaves like `umask` rather than `mode`, what to do ?

plugins/modules/mount.py
```diff
@@ -792,6 +803,16 @@ def main():
...
         if not os.path.exists(name) and not module.check_mode:
+            old_umask = None
+            if umask is not None:
+                if not isinstance(umask, int):
+                    try:
+                        umask = int(umask, 8)
+                    except ValueError as e:
+                        module.fail_json(msg="umask must be an octal integer: %s" % (to_native(e)))
+                old_umask = os.umask(umask)
+                os.umask(umask)
```
quidame, line 814: what's the purpose of calling `os.umask(umask)` twice ?

quidame (changes requested):
please update the changelog fragment - at least the name of the new parameter :); also, I don't understand the consecutive calls to `os.umask(umask)`. All the rest looks good to me.

plugins/modules/mount.py
satken2: Sorry, it's my mistake.

satken2:
@quidame Thank you for your polite explanation and suggestion.
I have understood your intentions and changed my code to use `umask` instead of `mode`.

quidame (approved):

verdurin:
I think this will solve the problem I recently saw when creating `CephFS` mounts with `posix.mount`, so I will be happy to see it merged.

plugins/modules/mount.py
```diff
@@ -105,6 +105,14 @@
         the original file back if you somehow clobbered it incorrectly.
     type: bool
     default: no
+  umask:
+    description:
+      - The umask to set before creating new directory(ies) for the mount point.
+        If the mount point already exists, this parameter is not used.
+      - Note that after running this task and the device being successfully mounted,
+        the mode of the original directory will be hidden by the target device.
+    type: raw
```
bcoca, line 114: I would change this to a 'strict string' to avoid all the 'octal/int' issues as we had to go through with mode in other modules.

PR #212, opened 2021-06-21
use_ssh_args is not in code. ssh_args is used

fedelemantuano:
##### SUMMARY
The `use_ssh_args` is not used in the code, but it's in documentation. This issue can confuse the Ansible user.
The option `ssh_args` is in code and it works.
It's possible to use it:

```
ssh_args: '-o "User=ansible"'
```

This issue is in all Ansible versions.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
synchronize

### DIFF
--- a/plugins/modules/synchronize.py
+++ b/plugins/modules/synchronize.py
@@ -135,10 +135,10 @@ options:
         that does not match the inventory user, you should set this parameter to C(no).
     type: bool
     default: yes
-  use_ssh_args:
+  ssh_args:
     description:
-      - Use the ssh_args specified in ansible.cfg. Setting this to `yes` will also make `synchronize` use `ansible_ssh_common_args`.
-    type: bool
+      - Use custom connection options for ssh connection. For more details `man ssh_options`. (e.g. C(-o "User=ansible"))
+    type: str
     default: no
   ssh_connection_multiplexing:
     description:


plugins/modules/synchronize.py
```diff
@@ -135,10 +135,10 @@
         that does not match the inventory user, you should set this parameter to C(no).
     type: bool
     default: yes
-  use_ssh_args:
+  ssh_args:
     description:
-      - Use the ssh_args specified in ansible.cfg. Setting this to `yes` will also make `synchronize` use `ansible_ssh_common_args`.
```
Akasurde, line 140: This parameter is used in https://github.com/ansible-collections/ansible.posix/blob/a994a58b23f22269e91e3611772105296538fe09/plugins/action/synchronize.py#L208

Akasurde (changes requested):

plugins/modules/synchronize.py
fedelemantuano: I could not see it in `module` and I could not use it in my playbooks.
`ssh_args` it's not in documentation, but I used it with successful.
The same error is [here](https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/files/synchronize.py#L146).

funkyfuture:
i assume this PR can be closed as it is based on an oversight.

maxamillion:
I closed based on comments, if it needs to be reopened please feel free and we can continue the discussion.

PR #213, opened 2021-06-24
synchronize: quotes around arguments

flybyray:
fix quoting for specific cmd arguments

Fixes:
- #24
- #190 

ISSUE TYPE:
- Bugfix Pull Request

COMPONENT NAME:
  module: synchronize

PR #220, opened 2021-07-12
Issues/171/suppress synchronize verbose output

### LINKED ISSUE #171: Suppress syncronize verbose output
##### SUMMARY

The synchronize module enables rsync's verbose output. This can flood the output with tons of output. Currently the only option for controlling this is `no_log: true`. However this suppresses all output, including errors. A mechanism should be provided for suppressing the changes, but still allow showing errors.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

synchronize

mandar242:
##### SUMMARY

Adding rsync 'quiet' option to synchronize module.
Allows to suppress synchronize verbose output.

Resolves #171 
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Synchronize


plugins/modules/synchronize.py
```diff
@@ -606,6 +619,9 @@ def _write_password_to_pipe(proc):
     else:
         (rc, out, err) = module.run_command(cmd)
 
+    if quiet:
+        cmd.append('--quiet')
+
```
gravesm, line 624: This should be moved up to before the paths get appended. Right now, it's being added after the command has been run, so it won't take effect.

plugins/modules/synchronize.py
```diff
@@ -184,6 +184,11 @@
     type: bool
     default: yes
     version_added: '1.3.0'
+  quiet:
+    description:
+      - This specifies rsync quiet option which on yes/true suppresses the non-error messages
+    type: bool
+    default: no
```
gravesm, line 191:
```suggestion
    default: no
    version_added: '1.3.0'
```

gravesm (changes requested):
Thanks for the PR @mandar242!

gravesm:
Could you also add a changelog fragment: https://docs.ansible.com/ansible/latest/community/development_process.html#community-changelogs?

mandar242:
> Could you also add a changelog fragment: https://docs.ansible.com/ansible/latest/community/development_process.html#community-changelogs?

Added.

changelogs/fragments/220_synchronize_add_quiet_option.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- synchronize - add the ``quiet`` option to supress non-error messages (https://github.com/ansible-collections/ansible.posix/issues/171).
```
Akasurde, line 2:
```suggestion
- synchronize - add the ``quiet`` option to suppress non-error messages (https://github.com/ansible-collections/ansible.posix/issues/171).
```

plugins/modules/synchronize.py
```diff
@@ -184,6 +184,12 @@
     type: bool
     default: yes
     version_added: '1.3.0'
+  quiet:
+    description:
+      - This specifies rsync quiet option which on yes/true suppresses the non-error messages
```
Akasurde, line 189:
```suggestion
      - This specifies rsync quiet option which on yes/true suppresses the non-error messages.
```

plugins/modules/synchronize.py
```diff
@@ -417,6 +429,7 @@ def main():
             delay_updates=dict(type='bool', default=True),
             mode=dict(type='str', default='push', choices=['pull', 'push']),
             link_dest=dict(type='list', elements='str'),
+            quiet=dict(type='bool',default=False)
```
Akasurde, line 432:
```suggestion
            quiet=dict(type='bool', default=False)
```

Akasurde (changes requested):
Small nit-picks. Could you please add tests for this change? Thanks.

Akasurde:
cc @quidame @saito-hideki Could you please review this? Thanks in advance.

saito-hideki (approved):
Hi! thank you for sending the PR!
Currently, we need to append `--allow-disabled` to perform the integration tests, but this PR and the integration tests worked fine in my testing environment.
BTW, a fix for the `aliases` file to enable the integration tests has been already included in another PR #213. So this PR looks reasonable to me even if it does not include the `aliases` file changing :)

gravesm:
This change actually may not work as is. It's breaking the ability to determine changed status. Testing locally for me shows no change when there is, in fact, a change. I think I'm not sure what extraneous output the OP was referring to is. The only case I can find where `-v` is being used is when `link_dest` is set, and the note in the code seems to suggest this is required here. @phemmer can you clarify what output you are referring to and possibly provide an example task to reproduce?

@mandar242 pending further information about what's being requested in this feature, we should add a few more tests to this. At least one to check that a task's result has changed and then a following one running the same task to check that the result has not changed.

quidame:
as reported by @gravesm , adding `--quiet` option just breaks `check_mode`. As said in *rsync(1)* manpage:
```
-q, --quiet                 suppress non-error messages
```
The `check_mode` is based on `--out-format` option (with a marker in), which is suppressed by the use of `--quiet`.

saito-hideki (commented):
Sorry guys, I missed the point pointed out by  @gravesm and @quidame. I'll revert my approval and wait for updates

quidame:
note that the `--out-format` option is given a marker (`<<CHANGED>>`), which is removed from the output before sending this output in the results (msg, stdout_lines). At this step (the module post-processing rsync results, and being aware of possible changes), I think it is doable to just keep rsync's stderr and suppress rsync's stdout if a user requests that with a `quiet=true` new module option :)

Maybe keep transfer metadata (number of files, duration...) anyway, even when hidding all the details, could be a good point.

quidame:
@mandar242 , I suggest you try with something like the following, which shouldn't break anything
```diff
@@ -619,14 +627,20 @@ def main():
     out_lines = out_clean.split('\n')
     while '' in out_lines:
         out_lines.remove('')
+
+    result = dict(changed=changed, rc=rc, cmd=cmdstr)
+
+    if quiet:
+        result['msg'] = "OUTPUT IS HIDDEN DUE TO 'quiet=true'"
+        result['stdout_lines'] = []
+    else:
+        result['msg'] = out_clean
+        result['stdout_lines'] = out_lines
+
     if module._diff:
-        diff = {'prepared': out_clean}
-        return module.exit_json(changed=changed, msg=out_clean,
-                                rc=rc, cmd=cmdstr, stdout_lines=out_lines,
-                                diff=diff)
+        result['diff'] = {'prepared': out_clean}
 
-    return module.exit_json(changed=changed, msg=out_clean,
-                            rc=rc, cmd=cmdstr, stdout_lines=out_lines)
+    return module.exit_json(**result)
 
 
 if __name__ == '__main__':
```

plugins/modules/synchronize.py
```diff
@@ -636,15 +633,20 @@ def _write_password_to_pipe(proc):
...
-        return module.exit_json(changed=changed, msg=out_clean,
-                                rc=rc, cmd=cmdstr, stdout_lines=out_lines,
-                                diff=diff)
 
-    return module.exit_json(changed=changed, msg=out_clean,
-                            rc=rc, cmd=cmdstr, stdout_lines=out_lines)
+    result = dict(changed=changed, rc=rc, cmd=cmdstr)
+
+    if quiet:
+        result['msg'] = "OUTPUT IS HIDDEN DUE TO 'quiet=true'"
```
phemmer, line 640: I think it'd be better to just omit the output, or output a summary ("X files synchronized"). An "OUTPUT IS HIDDEN" message feels a big aggressive (the caps), and unnecessary (I told it silent, I don't need it telling me I told it to do something).

plugins/modules/synchronize.py
```diff
@@ -636,15 +633,20 @@ def _write_password_to_pipe(proc):
...
+        result['stdout_lines'] = []
```
phemmer, line 641: I don't think this should be suppressed. `stdout_lines` doesn't get shown in the output, and having the data could be useful if someone wants to `register` the data into a variable and work with it. The problem described in #171 is just the message flooding the output.

plugins/modules/synchronize.py
```diff
@@ -619,15 +633,20 @@ def _write_password_to_pipe(proc):
...
+    else:
+        result['msg'] = out_clean
+        result['std_out'] = out_lines
```
quidame, line 644: based on @phemmer feedback, this could rather be:
```suggestion
    result = dict(changed=changed, rc=rc, cmd=cmdstr, stdout_lines=out_lines, msg=out_clean)
    if quiet:
        result['msg'] = "%s files/directories have been synchronized" % len(out_lines)
```

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
...
+- name: synchronize files with quiet option
+  synchronize:
+    src: '{{output_dir}}/foo.txt'
+    dest: '{{output_dir}}/foo.result'
+    quiet: true 
+  register: sync_result
+  ignore_errors: true
+- assert:
+    that:
+    - '''--quiet'' in sync_result.cmd'
```
quidame, line 239: this is not the case :)

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
...
+  - bar.result
+- name: synchronize files without quiet option 
+  synchronize:
+    src: '{{output_dir}}/foo.txt'
+    dest: '{{output_dir}}/foo.result'
+  register: sync_result
+  ignore_errors: true
+- assert:
+    that:
+    - '''--quiet'' not in sync_result.cmd'
```
quidame, line 255: this is not the case here too

quidame (changes requested):
@mandar242 , please take care to not break the existing interface. In https://github.com/ansible-collections/ansible.posix/pull/220/commits/f5088453d27ab739afa549717d7e3fc837346f6c you changed the `stdout_lines` result into `std_out`, please don't do that.

plugins/modules/synchronize.py
mandar242: This seems better, though I was wondering if len(out_lines) would give the actual number of files/directories syncd?

plugins/modules/synchronize.py
mandar242: Makes sense, the caps do seem aggressive. Thank you for the feedback.

plugins/modules/synchronize.py
mandar242: Would something like this work?

```
localhost | CHANGED => { 
    "changed": true,
    "cmd": "/usr/bin/rsync --delay-updates -F --compress --archive --out-format=<<CHANGED>>%i %n%L /home/tmp/src/path/ /home/tmp/dest/path/",
    "msg": "4 files/directories have been synchronized",
    "rc": 0,
    "stdout_lines": [
        ".d..t...... ./",
        ">f+++++++++ 1.txt",
        ">f+++++++++ 2.txt",
        ">f+++++++++ 3.txt",
        ">f+++++++++ 4.txt"
    ]
}
```

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
...
+  - bar.result
+- name: synchronize files without quiet option 
+  synchronize:
+    src: '{{output_dir}}/foo.txt'
+    dest: '{{output_dir}}/foo.result'
+  register: sync_result
+  ignore_errors: true
+- assert:
+    that:
+    - '''files/directories have been synchronized'' not in sync_result.msg'
```
quidame, line 255: This doesn't really make sense for me: despite this assertion, we have no certainty that this string is not in the result `msg` when the `synchronize` module succeeds, because we have no certainty that the `synchronize` module has succeeded, from a *logic* point of view.

Unless one *wants* to trigger an error, for example to test error handling of the module and ensure it behaves correctly under stress or with bad parameter values - and in that case we also *want* the playbook to not fail if a given task **fails as expected** - please don't use `ignore_errors=true`. Never. This makes just the tests passing even if the module is broken, that is ***exactly*** what the tests are made to avoid. An example of consistent use of `ignore_errors`: https://github.com/ansible-collections/community.general/blob/main/tests/integration/targets/filesize/tasks/errors.yml

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
   - directory_a/foo.txt
   - directory_a
   - directory_b
+- name: synchronize files with quiet option
+  synchronize:
+    src: '{{output_dir}}/foo.txt'
```
quidame, line 232: by default, some linters may complain about readability :)
```suggestion
    src: '{{ output_dir }}/foo.txt'
```

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
...
+  register: sync_result
+  ignore_errors: true
+- assert:
+    that:
+    - '''files/directories have been synchronized'' in sync_result.msg'
+- name: Cleanup
+  file:
+    state: absent
+    path: '{{output_dir}}/{{item}}'
+  with_items:
```
quidame, line 244: this begins to become more than deprecated...
```suggestion
  loop:
```

plugins/modules/synchronize.py
```diff
@@ -619,14 +632,17 @@ def _write_password_to_pipe(proc):
     out_lines = out_clean.split('\n')
     while '' in out_lines:
         out_lines.remove('')
+
+    result = dict(changed=changed, rc=rc, cmd=cmdstr, stdout_lines=out_lines, msg=out_clean)
+
+    if quiet:
+        changes = len(out_lines) - 1 if len(out_lines) >= 1 else 0
```
quidame, line 639: for better accuracy and consistency, this should be computed from the `changed` and `out` values, and the `changed_marker`, whose it's the purpose :) Maybe something as simple as
```suggestion
        changes = out.count(changed_marker) if changed else 0
```

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
   - directory_a/foo.txt
   - directory_a
   - directory_b
+- name: synchronize files with quiet option
+  synchronize:
+    src: '{{ output_dir }}/foo.txt'
+    dest: '{{ output_dir }}/foo.result'
+    quiet: true 
+  register: sync_result
+  ignore_errors: true
```
quidame, line 236: `ignore_errors=true` is for tasks that are expected to fail without failing the entire playbook/role. More precisely, in the case of a test role, this is for tasks that are not expected to succeed. It could be used, for example, to assert that when the module fails, it doesn't fail by raising an unhandled exception. I guess this task is not expected to fail, is it ?
```suggestion
```

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
...
+    path: '{{ output_dir }}/{{ item }}'
+  loop:
+  - foo.result
+  - bar.result
+- name: synchronize files without quiet option 
+  synchronize:
+    src: '{{ output_dir }}/foo.txt'
+    dest: '{{ output_dir }}/foo.result'
+  register: sync_result
+  ignore_errors: true
```
quidame, line 252:
```suggestion
```

tests/integration/targets/synchronize/tasks/main.yml
```diff
@@ -227,3 +227,36 @@
...
+- assert:
+    that:
+    - '''files/directories have been synchronized'' not in sync_result.msg'
+- name: Cleanup
+  file:
+    state: absent
+    path: '{{ output_dir }}/{{ item }}'
+  with_items:
```
quidame, line 260:
```suggestion
  loop:
```

quidame (approved):
LGTM

gravesm (approved):

saito-hideki (approved):

phemmer:
2 years later, what's needed for this to be merged? I see there's a change requested to add tests, but it looks like tests are present.

phemmer:
Ping? @Akasurde ?

Akasurde:
@phemmer sorry, but I am not actively looking into `ansible.posix` collection. @mandar242 Can you please rebase the branch and update `version_added`?

cc @saito-hideki @maxamillion Could you please take care of this PR? Thanks in advance.

phemmer:
@saito-hideki @maxamillion

Looks like this was rebased as requested several weeks ago. Is there any way to get this moving? Thanks

phemmer:
@saito-hideki @maxamillion

Hello???

phemmer:
@gravesm
@saito-hideki
@quidame
@Akasurde
@jillr 
@maxamillion 
@mrtwnklr 
@bachradsusi
@s-hertel 
@mib1185 
@cwchristerw 
@felixfontein 

Is this project maintained?

cwchristerw:
> @gravesm @saito-hideki @quidame @Akasurde @jillr @maxamillion @mrtwnklr @bachradsusi @s-hertel @mib1185 @cwchristerw @felixfontein
> 
> Is this project maintained?

Yes it is

plugins/modules/synchronize.py
```diff
@@ -186,6 +185,12 @@
     type: bool
     default: true
     version_added: '1.3.0'
+  quiet:
+    description:
+      - This option specifies quiet option which on true suppresses the output.
+    type: bool
+    default: no
+    version_added: '1.3.0'
```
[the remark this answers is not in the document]
felixfontein: It's 1.6.0, not 1.6.x :) But yes, that will be the next minor version, and the next one that will get released. (The only question is when..)

plugins/modules/synchronize.py
```diff
@@ -343,6 +348,12 @@
         src: /tmp/localpath/
         dest: /tmp/remotepath
         rsync_path: /usr/gnu/bin/rsync
+
+- name: Synchronization with quiet option enabled
+  ansible.posix.synchronize:
+    src: some/relative/path
+    dest: /some/absolute/path
+    quiet: yes
```
felixfontein, line 356:
```suggestion
    quiet: true
```

plugins/modules/synchronize.py
```diff
@@ -186,6 +185,12 @@
     type: bool
     default: true
     version_added: '1.3.0'
+  quiet:
+    description:
+      - This option specifies quiet option which on true suppresses the output.
+    type: bool
+    default: no
```
felixfontein, line 192:
```suggestion
    default: false
```

felixfontein (changes requested):
I'm not sure that this option makes sense, since `--quiet` means "suppress non-error messages". But the module **needs** the output to determine whether something changed.

So IMO this feature breaks the module and should not be merged.

felixfontein:
It seems that the rebase 10 months ago reintroduced passing `--quiet` to rsync (the problem has been discussed and fixed before that rebase).

phemmer:
@mandar242 since it has been years, if you no longer have the interest to maintain this PR, let me know and I can pick it up and open a new one.

mandar242:
> @mandar242 since it has been years, if you no longer have the interest to maintain this PR, let me know and I can pick it up and open a new one.

@phemmer I am fine with either. But given the number of commits made over the time, I think it will be easier and quicker to open new PR with changes from this pr. Rebase could be complex.

plugins/modules/synchronize.py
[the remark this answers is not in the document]
mandar242: added to 1.6.0

mandar242:
> > @mandar242 since it has been years, if you no longer have the interest to maintain this PR, let me know and I can pick it up and open a new one.
> 
> @phemmer I am fine with either. But given the number of commits made over the time, I think it will be easier and quicker to open new PR with changes from this pr. Rebase could be complex.

@phemmer the PR has been rebased successfully. waiting for reviews
cc @maxamillion @saito-hideki @felixfontein

plugins/modules/synchronize.py
```diff
@@ -601,6 +614,9 @@ def main():
 
     cmd.append(shlex_quote(source))
     cmd.append(shlex_quote(dest))
+    if quiet:
+        cmd.append('--quiet')
+
```
phemmer, line 619: This break the change detection as @felixfontein was mentioning [here](https://github.com/ansible-collections/ansible.posix/pull/220#issuecomment-2155447386)

Plus we probably want to keep the changes in `stdout_lines`. It's only `msg` that we want to suppress them from.

PR #218, opened 2021-07-02
Fix: Resolve CI checks fail - Module at add chdir(#102)

mandar242:
##### SUMMARY

Resolves CI checks failing at #102 which fixes #13 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

at

##### ADDITIONAL INFORMATION

```paste below


```

PR #228, opened 2021-07-23
Fixed job delete logic to cover RHEL6, CentOS6, *BSD and Solaris

saito-hideki:
##### SUMMARY

* `at -r` command-line option does not work on RHEL6 and CentOS6. Therefore, we will need to modify delete logic to use `atrm` command instead of `at -r`.
* Fixed search logic for job_id to perform correctly on *BSD and Solaris.

##### ISSUE TYPE
- Bugfix Pull Request
- Fixes #218 (CI tests issue with CentOS6)
 
##### COMPONENT NAME
- ansible.posix.at

##### ADDITIONAL INFORMATION
* RHEL6 and CentOS6 does not have `at -r` option:
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 6.10 (Santiago)
$ at -r
at: invalid option -- 'r'
Usage: at [-V] [-q x] [-f file] [-mMlbv] timespec ...
       at [-V] [-q x] [-f file] [-mMlbv] -t time
       at -c job ...
       atq [-V] [-q x]
       at [ -rd ] job ...
       atrm [-V] job ...
       batch
```
* The order of the job id number in the output column of `atq` on FreeBSD are different from Linux. We need to take care of the difference of job id position.
```
FreeBSD and NetBSD: job id is located at the end of the columns.
~~~
Date				Owner		Queue	Job#
Wed Dec 31 00:00:00 UTC 2200	root            c	9
~~~

OpenBSD: job id is located in the middle of the columns.
~~~
   Rank     Execution Date     Owner          Job       Queue
  1st   Jan  1, 2022 00:00   root       1640962800.c   c
~~~

Solaris: job id is located in the middle of the columns
~~~
 Rank     Execution Date     Owner      Job            Queue   Job Name
  1st   Jan  1, 2022 00:00   root       1640962800.a     a     stdin
~~~

Linux: job id is located at the beginning of the columns.
~~~
26	2022-01-01 00:00 a root
~~~
```
* `at -c` is different behavior on Solaris. So we need to read /var/spool/cron/atjobs/<job_id> to confirm the job details instead of `at -c` on Solaris nodes.

### DIFF
--- a/plugins/modules/at.py
+++ b/plugins/modules/at.py
@@ -86,9 +86,9 @@ def add_job(module, result, at_cmd, count, units, command, script_file):
     result['changed'] = True
 
 
-def delete_job(module, result, at_cmd, command, script_file):
+def delete_job(module, result, at_cmd, atrm_cmd, command, script_file):
     for matching_job in get_matching_jobs(module, at_cmd, script_file):
-        at_command = "%s -r %s" % (at_cmd, matching_job)
+        at_command = "%s %s" % (atrm_cmd, matching_job)
         rc, out, err = module.run_command(at_command, check_rc=True)
         result['changed'] = True
     if command:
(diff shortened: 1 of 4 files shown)


plugins/modules/at.py
```diff
@@ -151,6 +157,7 @@ def main():
     )
 
     at_cmd = module.get_bin_path('at', True)
+    atrm_cmd = module.get_bin_path('atrm', True)
```
Akasurde, line 160: Does `atrm` already exist on newer CentOS and RHEL machines?

saito-hideki: Hi thanks for the review!
Yeah, I have confirmed the following version has atrm command:
* RHEL6.10
* RHEL7.9
* RHEL8.4
* FreeBSD 13.0-STABLE
* NetBSD  9.1
* OpenBSD 6.9
* Solaris 11.4.0.15.0

saito-hideki:
It seems necessary to consider the behavior of get_matching_jobs () for Solaris. Therefore, I'll revert this PR to WIP.
https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/at.py#L99

saito-hideki:

saito-hideki:
@Akasurde is there any chance to review this PR? I don't know how many use cases there are in CentOS 6, but I think it's good that delete logic works properly :)

PR #234, opened 2021-07-29
Fix no enabling Icmp block inversion and masquerade

vrindle:
##### SUMMARY

Currently if you have icmp-block-inversion or masquerade set to any string including no they will be set. This PR fixes this behavior.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Firewalld
##### ADDITIONAL INFORMATION

```paste below

```


saito-hideki:
@vrindle Hello, thank you for the PR!
Modifying module code usually requires a changelog fragments file that describes the changes.  Looking at the output of CI tests, it has failed with the following error:

https://dashboard.zuul.ansible.com/t/ansible/build/18679f3c89b94077962aaf12cc3a3dcf
Your pull-request is missing a changelog fragment, please add one. It should explain to end users the reason for your change.
In this case, you need to create a changelog fragments file under changelogs/fragments directory according to the following document:

https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs
Example: https://github.com/ansible-collections/ansible.posix/blob/main/changelogs/fragments/247_firewalld.yml
Therefore, is it possible to create and add a fragments file and push those changes again?

plugins/modules/firewalld.py
```diff
@@ -845,7 +847,7 @@ def main():
         if changed is True:
             msgs.append("Changed icmp-block %s to %s" % (icmp_block, desired_state))
 
-    if icmp_block_inversion is not None:
+    if icmp_block_inversion not in NEGATIVES:
```
saito-hideki, line 850: In this conditional branch, if `icmp-block-inversion` is `no`, the icmp-block-inversion setting seems to be skipped.
Shouldn't this process be skipped only if icmp_block_inversion is undefined?

plugins/modules/firewalld.py
```diff
@@ -972,7 +974,7 @@ def main():
         changed, transaction_msgs = transaction.run()
         msgs = msgs + transaction_msgs
 
-    if masquerade is not None:
+    if masquerade not in NEGATIVES:
```
saito-hideki, line 977: This is a similar with the icmp_block_inversion review comment.

saito-hideki (changes requested):
Thank you for the PR! :)
I have left two comments with the change requests, also added the changelog fragments file to this PR.

vrindle:
Hi @saito-hideki
I think we should close this pr out, because we are eventually looking to create a second version of the ansible posix firewalld module that will end up using the module in the firewall system role. In that module we have already gotten the correct behavior for icmp-block-inversion as well as masquerading. We can leave the current state as is, so to not break functionality for existing customers. With your approval, I can go ahead and close this pull request now :)

saito-hideki:
@vrindle Thank you for your comment. I agree with your opinion. :)
We can close this case. I think that the solve this matter in the next version is a better approach that doesn't affect compatibility.

Issue #84, opened 2020-08-16
[mount] Ability to temporarily mount a filesystem without altering /etc/fstab


darkdragon-001:
##### SUMMARY
Migration of ansible/ansible#48134:

Ability to perform temporary mounts without altering `/etc/fstab`. Currently anything done with the mount module needs to manipulate `/etc/fstab`.

Additional parameter is needed to perform temporary mounts, such as with an ISO image that would be unmounted later in a play. Currently this can't be done with the mount module, as altering the `/etc/fstab` is required. Maybe call this additional parameter `modify_fstab` of type `bool`. Currently the only way to achieve this is via the `command` module using direct mount/unmount commands.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mount, fstab

##### ADDITIONAL INFORMATION

To mount an ISO image temporarily for pulling in data, but not make an entry in `/etc/fstab`.
```yaml
- name: Mount an ISO Temporarily
  mount:
    src: /path/to/my/iso.iso
    path: /path/to/mount/my/iso
    fstype: iso9660
    opts: ro
    modify_fstab: no
    state: mounted

- name: Unmount an ISO
  mount:
    path: /path/to/unmount/my/iso
    modify_fstab: no
    state: unmounted
```

chrystalis:
An additional use case discussed in the issues and PRs opened previously ([2571](https://github.com/ansible/ansible-modules-core/issues/2571), [19820](https://github.com/ansible/ansible/issues/19820), ([48134](https://github.com/ansible/ansible/issues/48134))) over the last 5 years, and one I am currently facing, is the need to **temporarily mount a filesystem that is already configured in fstab**.  Currently the mounted state requires that 'src' be provided, which is reasonable if updating fstab is required, but should be optional if a valid entry already exists in fstab.  

For example, if I need to add filesystems to a server and migrate existing content without changing the directory layout, I might have a playbook that creates the filesystems and mountpoints in their final desired state but does not actually mount them yet. Then I might have a playbook to mount them in a temporary location, migrate existing content to the new filesystems, and then unmount the temporaries and remount them in their new, permanent location.

NathanDotTo:
See https://github.com/ansible-collections/ansible.posix/issues/107.

quidame:
[...]

PR #264, opened 2021-09-12
Add ephemeral state to mount fs without altering fstab

NeodymiumFerBore:
##### SUMMARY

Add `ephemeral` possible value for `state` parameter.

The `ephemeral` state allows end-users to mount a volume on a given path, without altering an `fstab` file or creating a dummy one.

There have been debates about splitting this module into an `fstab` module and a `mount` module, but nothing has been done in 5 years. This is why I'd like to propose this feature.

**Downside**: the way the `posix.mount` module handles the mount options prevent it to be able to check exactly if the given `opts` perfectly match the mount options of an already mounted volume. To achieve this, the module would have to be aware of every `mount` default options, for all platforms. This is why the `ephemeral` state always return `changed=yes`. In other terms, a `remount` will always be triggered if the volume is already mounted, even if the options look to be the same. However, using *state* `unmounted` on a volume previously mounted with `ephemeral` behaves correctly.

##### ISSUE TYPE

- Feature Pull Request

Related issues:
-  [ansible/ansible#48134](https://github.com/ansible/ansible/issues/48134)
- #84 

##### COMPONENT NAME

mount

##### ADDITIONAL INFORMATION

**Example use case**

Sometimes it is handy to be able to temporarily mount a volume. I've seen this in some companies where they use Ansible to generate reports and put it on network shares. However, some admins don't look into mount options such as krb5 and multiuser for SMB shares. Being forced to use fstab-based mounts leads to clear text passwords being stored more or less temporarily on the host filesystem, requiring "manual" deletion (with the hassle of using blocks, rescues, always, etc.). This feature respond to this use case by providing a way to mount a volume without having to alter an fstab file.

**Description of changes**

- Edit DOCUMENTATION section to add ephemeral state
- Edit EXAMPLES section to add ephemeral state example
- Add new function `_set_ephemeral_args` to use instead of `_set_fstab_args` when using ephemeral state
- Add new function `_is_same_mount_src` to determine if the mounted volume on the destination path has the same source than the one supplied to the module
- Add new function `_get_mount_info` to avoid redundant code between functions `get_linux_mounts` and `_is_same_mount_src`
- Modify `get_linux_mount` to use the new function `_get_mount_info`. Behavior is preserved.
- Integrate `ephemeral` parameter treatment into `mounted` treatment, and add `if` statements to avoid IO from/to fstab
- Add `ephemeral` as a possible value for the `state` parameter in `main()`
- Add `required_if` dependencies for `ephemeral` state

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -31,12 +31,12 @@ options:
   src:
     description:
       - Device (or NFS volume, or something else) to be mounted on I(path).
-      - Required when I(state) set to C(present) or C(mounted).
+      - Required when I(state) set to C(present), C(mounted) or C(ephemeral).
     type: path
   fstype:
     description:
       - Filesystem type.
-      - Required when I(state) is C(present) or C(mounted).
+      - Required when I(state) is C(present), C(mounted) or C(ephemeral).
     type: str
   opts:
     description:
@@ -48,7 +48,7 @@ options:
       - Note that if set to C(null) and I(state) set to C(present),
         it will cease to work and duplicate entries will be made
         with subsequent runs.
-      - Has no effect on Solaris systems.
+      - Has no effect on Solaris systems or when used with C(ephemeral).
     type: str
     default: 0
   passno:
@@ -57,7 +57,7 @@ options:
       - Note that if set to C(null) and I(state) set to C(present),
         it will cease to work and duplicate entries will be made
         with subsequent runs.
-      - Deprecated on Solaris systems.
+      - Deprecated on Solaris systems. Has no effect when used with C(ephemeral).
     type: str
     default: 0
   state:
@@ -68,6 +68,13 @@ options:
       - If C(unmounted), the device will be unmounted without changing I(fstab).
       - C(present) only specifies that the device is to be configured in
         I(fstab) and does not trigger or require a mount.
+      - C(ephemeral) only specifies that the device is to be mounted, without changing
+        I(fstab). If it is already mounted, a remount will be triggered.
+        This will always return changed=True. If the mount point C(path)
+        has already a device mounted on, and its I(src) is different than C(src),
+        the module will fail to avoid unexpected unmount or mount point override.
+        If the mount point is not present, the mount point will be created.
+        The value of C(fstab) is ignored.
       - C(absent) specifies that the device mount's entry will be removed from
         I(fstab) and will also unmount the device and remove the mount
         point.
@@ -80,7 +87,7 @@ options:
         instead to work around this issue.
     type: str
     required: true
-    choices: [ absent, mounted, present, unmounted, remounted ]
+    choices: [ absent, mounted, present, unmounted, remounted, ephemeral ]
   fstab:
     description:
       - File to use instead of C(/etc/fstab).
@@ -89,6 +96,7 @@ options:
       - OpenBSD does not allow specifying alternate fstab files with mount so do not
         use this on OpenBSD with any state that operates on the live filesystem.
       - This parameter defaults to /etc/fstab or /etc/vfstab on Solaris.
+      - This parameter is ignored when I(state) is set to C(ephemeral).
     type: str
   boot:
     description:
@@ -100,6 +108,7 @@ options:
         to mount options in I(/etc/fstab).
       - To avoid mount option conflicts, if C(noauto) specified in C(opts),
         mount module will ignore C(boot).
+      - This parameter is ignored when I(state) is set to C(ephemeral).
     type: bool
     default: yes
   backup:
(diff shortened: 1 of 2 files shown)


plugins/modules/mount.py
```diff
@@ -671,12 +739,13 @@ def main():
             passno=dict(type='str', no_log=False),
             src=dict(type='path'),
             backup=dict(type='bool', default=False),
-            state=dict(type='str', required=True, choices=['absent', 'mounted', 'present', 'unmounted', 'remounted']),
+            state=dict(type='str', required=True, choices=['absent', 'mounted', 'present', 'unmounted', 'remounted', 'ephemeral']),
```
NeodymiumFerBore, line 742: `main()`: Add `ephemeral` as a possible value for the `state` parameter.

plugins/modules/mount.py
```diff
@@ -671,12 +739,13 @@ def main():
...
         ),
         supports_check_mode=True,
         required_if=(
             ['state', 'mounted', ['src', 'fstype']],
             ['state', 'present', ['src', 'fstype']],
+            ['state', 'ephemeral', ['src', 'fstype']]
```
NeodymiumFerBore, line 748: `main()`: Add `required_if` dependencies for `ephemeral` state.

plugins/modules/mount.py
```diff
@@ -747,15 +816,17 @@ def main():
...
-        if not os.path.exists(os.path.dirname(args['fstab'])):
-            os.makedirs(os.path.dirname(args['fstab']))
-        try:
-            open(args['fstab'], 'a').close()
-        except PermissionError as e:
-            module.fail_json(msg="Failed to open %s due to permission issue" % args['fstab'])
-        except Exception as e:
-            module.fail_json(msg="Failed to open %s due to %s" % (args['fstab'], to_native(e)))
+    # If state is 'ephemeral', we do not need fstab file
+    if module.params['state'] != 'ephemeral':
```
NeodymiumFerBore, line 820: Only this line changed (add `if` statement and fix block indent). This code block is in charge of creating the provided *fstab* file. For `ephemeral` state, the *fstab* file does not have to be created.

plugins/modules/mount.py
```diff
@@ -797,7 +870,7 @@ def main():
                         msg="Error unmounting %s: %s" % (name, msg))
 
             changed = True
-    elif state == 'mounted':
+    elif state == 'mounted' or state == 'ephemeral':
```
NeodymiumFerBore, line 873: `ephemeral` treatment is based on the `mounted` logic, bypassing the *fstab* read and write.

plugins/modules/mount.py
```diff
@@ -825,7 +898,11 @@ def main():
                 module.fail_json(
                     msg="Error making dir %s: %s" % (name, to_native(e)))
 
-        name, backup_lines, changed = _set_mount_save_old(module, args)
+        # ephemeral: completely ignore fstab
+        if state != 'ephemeral':
+            name, backup_lines, changed = _set_mount_save_old(module, args)
+        else:
+            name, backup_lines, changed = args['name'], [], False
```
NeodymiumFerBore, line 905: The function `_set_mount_save_old` records the mount in the *fstab* file. This step must be bypassed for `ephemeral` state.

plugins/modules/mount.py
```diff
@@ -835,7 +912,25 @@ def main():
...
+                    changed = True
+                else:
+                    module.fail_json(
+                        msg=(
+                            'Ephemeral mount point is already mounted with a different '
+                            'source than the specified one. Failing in order to prevent an '
+                            'unwanted unmount or override operation. Try replacing this command with '
+                            'a "state: unmounted" followed by a "state: ephemeral", or use '
+                            'a different destination path.'))
+
```
NeodymiumFerBore, line 931: When *state* is `ephemeral` or `mounted`, it tests if the provided mountpoint has already a volume mounted on it.

In the case of `ephemeral`, there is no *fstab* file to provide an authoritative answer to the question "What should be mounted here". To avoid an unwanted unmount and a possible service outage, the module fails if the volume already mounted on the mountpoint is not the same as the one provided in *src* (this is checked with the new function `_is_same_mount_src`).

plugins/modules/mount.py
```diff
@@ -847,7 +942,8 @@ def main():
             # A non-working fstab entry may break the system at the reboot,
             # so undo all the changes if possible.
             try:
-                write_fstab(module, backup_lines, args['fstab'])
+                if state != 'ephemeral':
+                    write_fstab(module, backup_lines, args['fstab'])
```
NeodymiumFerBore, line 946: This piece of code is the restoration of the *fstab* file to its previous state when the module fails. When *state* is `ephemeral`, the *fstab* file is never modified, so this step must be skipped.

NeodymiumFerBore (commented):
Added comments on all modifications in `mount.py`. I will be happy to discuss any of these points.

NeodymiumFerBore:
Renaming the branch closed the pull request. I thought the PR would be relocated, but this only works the other way around... Sorry for that. New PR: #267

PR #267, opened 2021-09-14
Add ephemeral state to mount fs without altering fstab

NeodymiumFerBore:
##### SUMMARY

Add `ephemeral` possible value for *state* parameter.

The `ephemeral` *state* allows end-users to mount a volume on a given path, without altering an *fstab* file or creating a dummy one.

There have been debates about splitting this module into an `fstab` module and a `mount` module, but nothing has been done in 5 years. This is why I'd like to propose this feature.

**Downside**: the way the `posix.mount` module handles mount options prevents it to be able to check exactly if the given *opts* perfectly match the mount options of an already mounted volume. To achieve this, the module would have to be aware of every `mount` default options, for all platforms. This is why `state=ephemeral` always return `changed=yes`.
In other terms, a `remount` will always be triggered if the volume is already mounted, even if the options look to be the same. Using `state=unmounted` on a volume previously mounted with `ephemeral` behaves correctly.

##### ISSUE TYPE

- Feature Pull Request

Related issues:
- [ansible/ansible#48134](https://github.com/ansible/ansible/issues/48134)
- #84 

##### COMPONENT NAME

mount

##### ADDITIONAL INFORMATION

**Example use case**

Sometimes it is handy to be able to temporarily mount a volume. I've seen this in couple companies where Ansible is used to generate reports and put it on network shares. However, some admins don't look into mount options such as krb5 and multiuser for SMB shares. Being forced to use fstab-based mounts leads to clear text passwords being stored more or less temporarily on the host filesystem, requiring "manual" deletion (with the hassle of using blocks, rescues, always, etc.). This feature respond to this use case by providing a way to mount a volume without having to alter an fstab file.

**Description of changes**

- Edit DOCUMENTATION section to add `ephemeral` *state*
- Edit EXAMPLES section to add `ephemeral` *state* example
- Add new function `_set_ephemeral_args` to use instead of `_set_fstab_args` when using ephemeral state
- Add new function `_is_same_mount_src` to determine if the mounted volume on the destination path has the same source than the one supplied to the module
- Add new function `_get_mount_info` to avoid redundant code between functions `get_linux_mounts` and `_is_same_mount_src`
- Modify `get_linux_mount` to use the new function `_get_mount_info`. Original behavior is preserved.
- Integrate `ephemeral` parameter treatment into `mounted` treatment, and add `if` statements to avoid IO from/to fstab
- Add `ephemeral` as a possible value for the state parameter in `main()`
- Add `required_if` dependencies for `ephemeral` state

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -74,7 +74,7 @@ options:
         has already a device mounted on, and its source is different than I(src),
         the module will fail to avoid unexpected unmount or mount point override.
         If the mount point is not present, the mount point will be created.
-        The I(fstab) is completely ignored.
+        The I(fstab) is completely ignored. This option is added in version 1.5.0.
       - C(absent) specifies that the device mount's entry will be removed from
         I(fstab) and will also unmount the device and remove the mount
         point.
@@ -726,7 +726,7 @@ def _is_same_mount_src(module, src, mountpoint, linux_mounts):
         # the mountpoint is a bind mount AND the source FS is the same than 'src'.
         # is_bind_mounted() is not reliable on Solaris, NetBSD and OpenBSD.
         # But we can rely on 'mount -v' on all other platforms, and Linux non-bind mounts.
-        if (is_bind_mounted(module, linux_mounts, mountpoint, src)):
+        if is_bind_mounted(module, linux_mounts, mountpoint, src):
             return True
 
     # mount with parameter -v has a close behavior on Linux, *BSD, SunOS
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -31,12 +31,12 @@ options:
   src:
     description:
       - Device (or NFS volume, or something else) to be mounted on I(path).
-      - Required when I(state) set to C(present) or C(mounted).
+      - Required when I(state) set to C(present), C(mounted) or C(ephemeral).
     type: path
   fstype:
     description:
       - Filesystem type.
-      - Required when I(state) is C(present) or C(mounted).
+      - Required when I(state) is C(present), C(mounted) or C(ephemeral).
     type: str
   opts:
     description:
@@ -48,7 +48,7 @@ options:
       - Note that if set to C(null) and I(state) set to C(present),
         it will cease to work and duplicate entries will be made
         with subsequent runs.
-      - Has no effect on Solaris systems.
+      - Has no effect on Solaris systems or when used with C(ephemeral).
     type: str
     default: '0'
   passno:
@@ -57,7 +57,7 @@ options:
       - Note that if set to C(null) and I(state) set to C(present),
         it will cease to work and duplicate entries will be made
         with subsequent runs.
-      - Deprecated on Solaris systems.
+      - Deprecated on Solaris systems. Has no effect when used with C(ephemeral).
     type: str
     default: '0'
   state:
@@ -68,6 +68,13 @@ options:
       - If C(unmounted), the device will be unmounted without changing I(fstab).
       - C(present) only specifies that the device is to be configured in
         I(fstab) and does not trigger or require a mount.
+      - C(ephemeral) only specifies that the device is to be mounted, without changing
+        I(fstab). If it is already mounted, a remount will be triggered.
+        This will always return changed=True. If the mount point I(path)
+        has already a device mounted on, and its source is different than I(src),
+        the module will fail to avoid unexpected unmount or mount point override.
+        If the mount point is not present, the mount point will be created.
+        The I(fstab) is completely ignored.
       - C(absent) specifies that the device mount's entry will be removed from
         I(fstab) and will also unmount the device and remove the mount
         point.
@@ -77,10 +84,12 @@ options:
         applied to the remount, but will not change I(fstab).  Additionally,
         if I(opts) is set, and the remount command fails, the module will
         error to prevent unexpected mount changes.  Try using C(mounted)
-        instead to work around this issue.
+        instead to work around this issue.  C(remounted) expects the mount point
+        to be present in the I(fstab). To remount a mount point not registered
+        in I(fstab), use C(ephemeral) instead, especially with BSD nodes.
     type: str
     required: true
-    choices: [ absent, mounted, present, unmounted, remounted ]
+    choices: [ absent, mounted, present, unmounted, remounted, ephemeral ]
   fstab:
     description:
       - File to use instead of C(/etc/fstab).
@@ -89,6 +98,7 @@ options:
       - OpenBSD does not allow specifying alternate fstab files with mount so do not
         use this on OpenBSD with any state that operates on the live filesystem.
       - This parameter defaults to /etc/fstab or /etc/vfstab on Solaris.
+      - This parameter is ignored when I(state) is set to C(ephemeral).
     type: str
   boot:
     description:
@@ -100,6 +110,7 @@ options:
         to mount options in I(/etc/fstab).
       - To avoid mount option conflicts, if C(noauto) specified in C(opts),
         mount module will ignore C(boot).
+      - This parameter is ignored when I(state) is set to C(ephemeral).
     type: bool
     default: yes
   backup:
@@ -184,6 +195,14 @@ EXAMPLES = r'''
     boot: no
     state: mounted
     fstype: nfs
+
+- name: Mount ephemeral SMB volume
+  ansible.posix.mount:
+    src: //192.168.1.200/share
+    path: /mnt/smb_share
+    opts: "rw,vers=3,file_mode=0600,dir_mode=0700,dom={{ ad_domain }},username={{ ad_username }},password={{ ad_password }}"
+    fstype: cifs
+    state: ephemeral
 '''
 
 import errno
@@ -430,6 +449,24 @@ def _set_fstab_args(fstab_file):
     return result
 
 
+def _set_ephemeral_args(args):
+    result = []
+    # Set fstype switch according to platform. SunOS/Solaris use -F
+    if platform.system().lower() == 'sunos':
+        result.append('-F')
+    else:
+        result.append('-t')
+    result.append(args['fstype'])
+
+    # Even if '-o remount' is already set, specifying multiple -o is valid
+    if args['opts'] != 'defaults':
+        result += ['-o', args['opts']]
+
+    result.append(args['src'])
+
+    return result
+
+
 def mount(module, args):
     """Mount up a path or remount if needed."""
 
@@ -446,7 +483,11 @@ def mount(module, args):
                     'OpenBSD does not support alternate fstab files. Do not '
                     'specify the fstab parameter for OpenBSD hosts'))
     else:
-        cmd += _set_fstab_args(args['fstab'])
+        if module.params['state'] != 'ephemeral':
+            cmd += _set_fstab_args(args['fstab'])
+
+    if module.params['state'] == 'ephemeral':
+        cmd += _set_ephemeral_args(args)
 
     cmd += [name]
 
@@ -498,18 +539,24 @@ def remount(module, args):
                     'OpenBSD does not support alternate fstab files. Do not '
                     'specify the fstab parameter for OpenBSD hosts'))
     else:
-        cmd += _set_fstab_args(args['fstab'])
+        if module.params['state'] != 'ephemeral':
+            cmd += _set_fstab_args(args['fstab'])
+
+    if module.params['state'] == 'ephemeral':
+        cmd += _set_ephemeral_args(args)
 
     cmd += [args['name']]
     out = err = ''
 
     try:
-        if platform.system().lower().endswith('bsd'):
+        if module.params['state'] != 'ephemeral' and platform.system().lower().endswith('bsd'):
             # Note: Forcing BSDs to do umount/mount due to BSD remount not
             # working as expected (suspect bug in the BSD mount command)
             # Interested contributor could rework this to use mount options on
             # the CLI instead of relying on fstab
             # https://github.com/ansible/ansible-modules-core/issues/5591
+            # Note: this does not affect ephemeral state as all options
+            # are set on the CLI and fstab is expected to be ignored.
             rc = 1
         else:
             rc, out, err = module.run_command(cmd)
@@ -663,6 +710,47 @@ def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
     return mounts
 
 
+def _is_same_mount_src(module, src, mountpoint, linux_mounts):
+    """Return True if the mounted fs on mountpoint is the same source than src. Return False if mountpoint is not a mountpoint"""
+    # If the provided mountpoint is not a mountpoint, don't waste time
+    if (
+            not ismount(mountpoint) and
+            not is_bind_mounted(module, linux_mounts, mountpoint)):
+        return False
+
+    # Treat Linux bind mounts
+    if platform.system() == 'Linux' and linux_mounts is not None:
+        # For Linux bind mounts only: the mount command does not return
+        # the actual source for bind mounts, but the device of the source.
+        # is_bind_mounted() called with the 'src' parameter will return True if
+        # the mountpoint is a bind mount AND the source FS is the same than 'src'.
+        # is_bind_mounted() is not reliable on Solaris, NetBSD and OpenBSD.
+        # But we can rely on 'mount -v' on all other platforms, and Linux non-bind mounts.
+        if (is_bind_mounted(module, linux_mounts, mountpoint, src)):
+            return True
+
+    # mount with parameter -v has a close behavior on Linux, *BSD, SunOS
+    # Requires -v with SunOS. Without -v, source and destination are reversed
+    # Output format differs from a system to another, but field[0:3] are consistent: [src, 'on', dest]
+    cmd = '%s -v' % module.get_bin_path('mount', required=True)
+    rc, out, err = module.run_command(cmd)
+    mounts = []
+
+    if len(out):
+        mounts = to_native(out).strip().split('\n')
+    else:
+        module.fail_json(msg="Unable to retrieve mount info with command '%s'" % cmd)
+
+    for mnt in mounts:
+        fields = mnt.split()
+        mp_src = fields[0]
+        mp_dst = fields[2]
+        if mp_src == src and mp_dst == mountpoint:
+            return True
+
+    return False
+
+
 def main():
     module = AnsibleModule(
         argument_spec=dict(
@@ -675,12 +763,13 @@ def main():
             passno=dict(type='str', no_log=False, default='0'),
             src=dict(type='path'),
             backup=dict(type='bool', default=False),
-            state=dict(type='str', required=True, choices=['absent', 'mounted', 'present', 'unmounted', 'remounted']),
+            state=dict(type='str', required=True, choices=['absent', 'mounted', 'present', 'unmounted', 'remounted', 'ephemeral']),
         ),
         supports_check_mode=True,
         required_if=(
             ['state', 'mounted', ['src', 'fstype']],
             ['state', 'present', ['src', 'fstype']],
+            ['state', 'ephemeral', ['src', 'fstype']]
         ),
     )
 
@@ -751,15 +840,17 @@ def main():
 
     # If fstab file does not exist, we first need to create it. This mainly
     # happens when fstab option is passed to the module.
-    if not os.path.exists(args['fstab']):
-        if not os.path.exists(os.path.dirname(args['fstab'])):
-            os.makedirs(os.path.dirname(args['fstab']))
-        try:
-            open(args['fstab'], 'a').close()
-        except PermissionError as e:
-            module.fail_json(msg="Failed to open %s due to permission issue" % args['fstab'])
-        except Exception as e:
-            module.fail_json(msg="Failed to open %s due to %s" % (args['fstab'], to_native(e)))
+    # If state is 'ephemeral', we do not need fstab file
+    if module.params['state'] != 'ephemeral':
+        if not os.path.exists(args['fstab']):
+            if not os.path.exists(os.path.dirname(args['fstab'])):
+                os.makedirs(os.path.dirname(args['fstab']))
+            try:
+                open(args['fstab'], 'a').close()
+            except PermissionError as e:
+                module.fail_json(msg="Failed to open %s due to permission issue" % args['fstab'])
+            except Exception as e:
+                module.fail_json(msg="Failed to open %s due to %s" % (args['fstab'], to_native(e)))
 
     # absent:
     #   Remove from fstab and unmounted.
@@ -770,6 +861,8 @@ def main():
     # mounted:
     #   Add to fstab if not there and make sure it is mounted. If it has
     #   changed in fstab then remount it.
+    # ephemeral:
+    #   Do not change fstab state, but mount.
 
     state = module.params['state']
     name = module.params['path']
@@ -801,7 +894,7 @@ def main():
                         msg="Error unmounting %s: %s" % (name, msg))
 
             changed = True
-    elif state == 'mounted':
+    elif state == 'mounted' or state == 'ephemeral':
         dirs_created = []
         if not os.path.exists(name) and not module.check_mode:
             try:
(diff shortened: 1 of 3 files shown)


plugins/modules/mount.py
```diff
@@ -602,6 +642,17 @@ def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
...
+    return lines
+
+
+def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
+    """Gather mount information"""
+
+    lines = _get_mount_info(module, mntinfo_file)
+    # Keep same behavior than before
+    if lines is None:
+        return
```
NeodymiumFerBore, line 654: The function `get_linux_mount` used to implement directly the `_get_mount_info` logic. Since the new function `_is_same_mount_src` also needed the mount info, I centralized this process in `_get_mount_info` to avoid code redundancy.

The behavior of `get_linux_mounts` is unchanged (return nothing if `IOError` is raised when opening the file, fail module if `IOError` is raised when closing the file).

plugins/modules/mount.py
```diff
@@ -659,6 +710,25 @@ def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
     return mounts
 
 
+def _is_same_mount_src(module, args, mntinfo_file="/proc/self/mountinfo"):
+    """Return True if the mounted fs on mountpoint is the same source than src"""
```
NeodymiumFerBore, line 714: This function is used for the `ephemeral` *state*: if this function returns False, that means that the provided mount point has already a volume mounted on it, and it is NOT the volume given in *src*. This leads to a module failure, to avoid an unwanted unmount.

plugins/modules/mount.py
```diff
@@ -659,6 +710,25 @@ def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
...
+    lines = _get_mount_info(module)
+
+    # If this function is used and we cannot retrieve mount info, we must fail to avoid unexpected behavior
+    if lines is None:
+        module.fail_json(msg="Unable to retrieve mount info from '%s'" % mntinfo_file)
+
+    for line in lines:
+        fields = line.split()
+        if fields[4] == mountpoint and fields[-2] == src:
+            return True
```
NeodymiumFerBore, line 726: `fields[4]` contains a mountpoint and `fields[-2]` contains the associated source. The logic behind picking these fields was already implemented in `get_linux_mounts`, I only copy-pasted after test and review.

plugins/modules/mount.py
```diff
@@ -847,7 +944,8 @@ def main():
             # A non-working fstab entry may break the system at the reboot,
             # so undo all the changes if possible.
             try:
-                write_fstab(module, backup_lines, args['fstab'])
+                if state != 'ephemeral':
+                    write_fstab(module, backup_lines, args['fstab'])
```
NeodymiumFerBore, line 948: This piece of code is the restoration of the *fstab* file to its previous state when the module fails. When *state* is `ephemeral`, the *fstab* file is never modified, so this step must be skipped.

NeodymiumFerBore (commented):
Initial review: comments about major changes.

saito-hideki:
@NeodymiumFerBore Thank you for the pull request. If it is possible, can you create integration tests for the `ephemeral` option?
- https://github.com/ansible-collections/ansible.posix/blob/main/tests/integration/targets/mount/tasks/main.yml

NeodymiumFerBore:
Hello @saito-hideki . Thank you for your interest in my PR. Yes I can write integration tests. I don't have much time right now. I will push the tests into this branch.

frittentheke:
> Hello @saito-hideki . Thank you for your interest in my PR. Yes I can write integration tests. I don't have much time right now. I will push the tests into this branch.

@NeodymiumFerBore  thanks for this feature / option. Would be cool if you could write those test so this can be added to Ansible ;-)

NeodymiumFerBore:
Hello. Glad you appreciate this feature, thanks for the feedback :)
I had a lot of work and stuff going on lately, I haven't take the time to write it. Will try to do it this weekend!

NeodymiumFerBore:
Hello @saito-hideki . I started writing integration tests and got into some behavior review. Maybe this is intended, but this is not documented as far as I know.

I will keep writing integration tests while keeping the current behavior. If you have some time, please give me your opinion on the following. We may fix it with another PR.

## Weird behavior

As of now, the `absent` *state* does NOT unmount a mounted FS if it is already absent from *fstab*. [Link to the code](https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/mount.py#L778): the condition `if changed` is `True` only if changes occurred in the fstab file.

So, if someone mounts a FS with `ephemeral` and then tries to unmount it with `absent`, it will return `changed=false` and the FS will NOT be unmounted.

This behavior would be the same in the following situation (without ephemeral option):
- someone manually mounts a FS on a host on /mnt/my_mount, without Ansible, and without altering the fstab
- a playbook runs with `state=absent` and `path=/mnt/my_mount`
- the playbook returns `changed=false` and /mnt/my_mount is still mounted

## How to reproduce

```console
$ # Setup test environment
$ mkdir /tmp/my_mountpoint
$ dd if=/dev/zero of=/tmp/test.img bs=10M count=10
$ mkfs.ext4 /tmp/test.img

$ # Mount the file without adding it to fstab
$ sudo mount -t ext4 /tmp/test.img /tmp/my_mountpoint

$ # Take note of the current state
$ df -HT | grep my_mountpoint
/dev/loop1                                  ext4       98M   74k   90M   1% /tmp/my_mountpoint
$ grep my_mountpoint /etc/fstab
(no output)
$ sha1sum /etc/fstab
74528a180d8b9be29585862713736156ebdebbe0  /etc/fstab

$ # Play the mount module with state = absent
$ ansible --become -K -m mount -a 'path=/tmp/my_mountpoint state=absent' localhost
BECOME password: 
localhost | SUCCESS => {
    "backup_file": "",
    "boot": "yes",
    "changed": false,
    "dump": "0",
    "fstab": "/etc/fstab",
    "name": "/tmp/my_mountpoint",
    "opts": "defaults",
    "passno": "0"
}

$ # The system state did not change
$ df -HT | grep my_mountpoint
/dev/loop1                                  ext4       98M   74k   90M   1% /tmp/my_mountpoint
$ grep my_mountpoint /etc/fstab
(no output)
$ sha1sum /etc/fstab # no change occurred
74528a180d8b9be29585862713736156ebdebbe0  /etc/fstab

$ # Nuke the test environment
$ sudo umount /tmp/my_mountpoint
$ rm -rf /tmp/my_mountpoint
$ rm -f /tmp/test.img
```

## My questions to you

- Is this behavior expected? [I'm not sure](https://github.com/ansible-collections/ansible.posix/blob/792714f598fc4475c7ec007f6391cb4b312e764d/plugins/modules/mount.py#L71)
- Should I fix this behavior, so `absent` actually unmounts the path even if the fstab was not changed?
- Should I just add a note in the `absent` documentation about this fact? Like "A filesystem mounted on a mountpoint that is not registered in the fstab will not be unmounted with this option. Use unmount instead".

I was thinking about a fix like:
- if the mountpoint was present in the fstab, keep the current behavior:
  - remove entry in fstab
  - unmount the mountpoint
  - delete the mountpoint
- else if the mountpoint was not present in the fstab:
  - unmount the mountpoint is necessary (and trigger a change)
  - do NOT delete the mountpoint

What is your opinion about this?

NeodymiumFerBore:
Tests fail on RHEL 7.9 [remote devel](https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=35494&view=logs&j=d872f81b-b06d-5557-66bc-b224cd892747&t=65c93845-713b-572d-b453-29b5e62c4fef) and [remote 2.12](https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=35494&view=logs&j=577c1ae7-f904-5529-074c-8605bda5d5c1&t=7bf9051c-d445-5442-0ce2-844eacad923f): assertion `"'/tmp/myfs' is mount"` is not asserted, though it is asserted on RHEL 8. Any idea that would save me some time?
Will go into troubleshooting this next week.

NeodymiumFerBore:
Hello @saito-hideki . Should I squash my commits into a single one and force push my branch? Rebase it again on upstream main?

Is there any required action from me before the merge can occur?

plugins/modules/mount.py
saito-hideki: @NeodymiumFerBore I have tested this FIX and it worked first run on *BSD, Solars, and Linux.
However, when I tried the second run, it only succeeded on Linux host, and other OS has been failed with the following error:

```shell
fatal: [freebsd]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/local/bin/python3.8"}, "changed": false, "msg": "Unable to retrieve mount info from '/proc/self/mountinfo'"}
fatal: [netbsd]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/pkg/bin/python3.9"}, "changed": false, "msg": "Unable to retrieve mount info from '/proc/self/mountinfo'"}
fatal: [openbsd]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/local/bin/python3.8"}, "changed": false, "msg": "Unable to retrieve mount info from '/proc/self/mountinfo'"}
fatal: [solaris]: FAILED! => {"changed": false, "msg": "Unable to retrieve mount info from '/proc/self/mountinfo'"}
```

Even if the `ephemeral` option only works on Linux, I think we need to take care of other OS.

saito-hideki (changes requested):
I think we need to take care of how it works on non-Linux operating systems. For example, it outputs a warning and does not actually perform the ephemeral mount.

plugins/modules/mount.py
NeodymiumFerBore: ## Why integration tests fail on *BSD and Unix as of now

@saito-hideki I don't know what values you used for the *src* parameter, but as of now, it might success if you are using network sources (currently, the tests fail on *bsd and solaris due to the way they handle loop devices, I'm implementing a fix for that).

## Why manual tests might work first run, but not second

If it failed on second run (and I assume you didn't run the integration tests, but did manual tests), I think it is due to the function `_is_same_mount_src`. It uses the function `_get_mount_info` without checking if the system is Linux, and `/proc/self/mountinfo` can only be reliably used on Linux systems. As does `/sys/block/loop${x}/loop/backing_file`.

## Note about checking the backing file of loop devices

This might be useless to try to get the backing file of a loop device on some systems. When we format/mount a file, Linux manipulates this file a block device by **implicitly** creating/using a loop device.
Solaris does not support this, and we need to **explicitly** create a loop file interface for the file (with `lofiadm`). I have to do more tests about this for OpenBSD, FreeBSD and NetBSD.

In this version of my PR, when using `ephemeral` *state*, `mount.py` tries to verify if *path* is a mountpoint, and if it finds a loop device (`/dev/loop[0-9]+`), it tries to retrieve the backing file through `/sys/block/loop${x}/loop/backing_file`. This is required to make the `ephemeral` *state* compatible with mounting files (implicit loop devices).

On systems that do not implicitly creates this loop device (such as Solaris), **it might be useless to try to find the backing file of a loop device**, as if a user tries to mount a file, he will first manually find/create the loop device, as he would on a shell.

I working on this.

plugins/modules/mount.py
```diff
@@ -663,6 +716,41 @@ def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
     return mounts
 
 
+def _is_same_mount_src(module, args, mntinfo_file="/proc/self/mountinfo"):
+    """Return True if the mounted fs on mountpoint is the same source than src"""
+    mountpoint = args['name']
+    src = args['src']
+    lines = _get_mount_info(module)
```
NeodymiumFerBore, line 723: See [comment](https://github.com/ansible-collections/ansible.posix/pull/267/files/13916108fa32a53804c2e2698afe33565704b939#r830526629): this call (or the `_get_mount_info` function) must be system aware

plugins/modules/mount.py
NeodymiumFerBore: TLDR: Im implementing a fix so the `ephemeral` *state* works fine on *BSD, Solaris and Linux. It might take some time and a lot of testing.

NeodymiumFerBore:
Hello @saito-hideki . Is there a reachable NFS target in the test environment? If yes, what's its name?
It would highly simplify integration tests. Using loop devices on *BSD and Solaris is not transparent, and they all have different tools (`lofiadm`, `mdconfig`, `vnconfig`).
Integration tests in my last push work for (py2/py3) Solaris 11, NetBSD 8.2/9.2, FreeBSD 13, OpenBSD 6.9, but it's pretty heavy...
Thank you.

plugins/modules/mount.py
NeodymiumFerBore: Moved from `/proc/self/mountinfo` to `mount -v`, which has a close behavior on Linux, Solaris and *BSD systems.
Also implemented integration tests for Solaris/*BSD

tests/integration/targets/mount/tasks/main.yml
```diff
@@ -406,3 +420,283 @@
     - /tmp/myfs1
     - /tmp/test_fstab
   when: ansible_system in ('Linux')
+
+######################################################################################################
+######################################################################################################
+######################################################################################################
```
saito-hideki, line 426: It would be better to remove the three-line separator (#). Because test logic for the ephemeral option has already been separated by the subsequent block :)

tests/integration/targets/mount/tasks/main.yml
```diff
@@ -406,3 +420,283 @@
...
+  environment:
+    PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+  block:
+  - name: Patch centos repolist
+    ansible.builtin.shell:
+      cmd: "{{ item }}"
+    with_items:
+    - sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
+    - sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
+    when: ansible_distribution in ('CentOS') and ansible_distribution_version|int >= 8
```
saito-hideki, line 438: CentOS8(or later) is no longer supported for the integration tests. So I think you can remove this `Patch centos repolist` task from the integration test.

- https://github.com/ansible-collections/overview/issues/45#issuecomment-944657318

saito-hideki (changes requested):
Thank you for your contribution! There are two small changes requested for the integration test :)

saito-hideki:
Hi @NeodymiumFerBore 

> Integration tests in my last push work at least on (py2/py3) Solaris 11, NetBSD 8.2/9.2, FreeBSD 13, OpenBSD 6.9, but it's pretty heavy...
> Thank you.

Thank you for your hard work. Your integration tests worked not only Linux but also on Solaris and *BSD(my test environment is Solaris11.4, FreeBSD13, NetBSD9.2, and OpenBSD6.9).
I have left two trivial change requests and will set approve flag once it is fixed.

@Akasurde if you have B/W, is it possible to review this PR? if you have B/W, is it possible to review this PR? This PR looks pretty reasonable for me.

tests/integration/targets/mount/tasks/main.yml
NeodymiumFerBore: This will be removed in the next push (squashed). This was just eye-candy during development, sorry for that!

tests/integration/targets/mount/tasks/main.yml
NeodymiumFerBore: Removed in latest push

NeodymiumFerBore:
Hello @saito-hideki . Thank you for you warm feedback! :)) The requested changes have been addressed.

In my last push (squashed commits), the [CI failed](https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=40307&view=logs&j=81ddb6ae-3e40-537b-e820-0934407cab38&t=b5bc49be-44bf-5cf5-4e26-db7c7b8dbcdf&l=142) with error `ERROR! Error when getting collection version metadata for community.general:1.3.5 from default (https://galaxy.ansible.com/api/) (HTTP Code: 429, Message: Too Many Requests Code: Unknown)`. I don't think the code is responsible for this error.
Is there any way I can trigger the CI again without pushing dummy stuff?

saito-hideki:
Closing and reopening for CI trigger

saito-hideki:
@NeodymiumFerBore  Thank you for reporting the CI issue. The galaxy issue that you reported seems fixed(galaxy server API possibly unavailable at that time).
However, it seems that there's another problem on the CI tests side. This error seems to be a problem on the CI test side and not directly related to this PR.
So we will check CI testing and let you know if we have any clue.

saito-hideki:
Closing and reopening for CI trigger

saito-hideki:
Hi @NeodymiumFerBore , I have fixed one of the unit test issues, also test team seems fixed several issues regarding execution environment tests.
Therefore, if it is possible, can you rebase this PR with the latest main branch? Once you rebased it, I think this PR will pass the CI tests.

NeodymiumFerBore:
Rebased PR branch to master dc4da60affbe29e7706851ec73dab09bd11bdb44.

NeodymiumFerBore:
Hello @saito-hideki , thank you for your quick fix on the CI, build succeeded 🥳 Let me know if there is anything I should do for the merge to occur!

saito-hideki (approved):
Looks good to me :)

saito-hideki:
@AkasurdeI think this PR is reasonable. I would appreciate it if you could review this PR :)
Thanks!

plugins/modules/mount.py
```diff
@@ -663,6 +710,47 @@ def get_linux_mounts(module, mntinfo_file="/proc/self/mountinfo"):
...
+
+    # Treat Linux bind mounts
+    if platform.system() == 'Linux' and linux_mounts is not None:
+        # For Linux bind mounts only: the mount command does not return
+        # the actual source for bind mounts, but the device of the source.
+        # is_bind_mounted() called with the 'src' parameter will return True if
+        # the mountpoint is a bind mount AND the source FS is the same than 'src'.
+        # is_bind_mounted() is not reliable on Solaris, NetBSD and OpenBSD.
+        # But we can rely on 'mount -v' on all other platforms, and Linux non-bind mounts.
+        if (is_bind_mounted(module, linux_mounts, mountpoint, src)):
```
Akasurde, line 729:
```suggestion
        if is_bind_mounted(module, linux_mounts, mountpoint, src):
```

plugins/modules/mount.py
```diff
@@ -68,6 +68,13 @@
       - If C(unmounted), the device will be unmounted without changing I(fstab).
       - C(present) only specifies that the device is to be configured in
         I(fstab) and does not trigger or require a mount.
+      - C(ephemeral) only specifies that the device is to be mounted, without changing
+        I(fstab). If it is already mounted, a remount will be triggered.
+        This will always return changed=True. If the mount point I(path)
+        has already a device mounted on, and its source is different than I(src),
+        the module will fail to avoid unexpected unmount or mount point override.
+        If the mount point is not present, the mount point will be created.
+        The I(fstab) is completely ignored.
```
Akasurde, line 77:
```suggestion
        The I(fstab) is completely ignored. This option is added in version 1.4.0.
```

Akasurde (commented):
few nitpicks, rest LGTM.

NeodymiumFerBore:
Hi @Akasurde . Thank you for your review :) The ansible-galaxy API seems to have some ["Too many requests" issues](https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=42731&view=logs&j=2f5c63df-f099-552d-bc84-cb4b96c98a86&t=91da7c18-2e80-5bd6-c4b2-84ba36a5f091&l=144), and fails the CI.

I will close/re-open the PR to trigger the CI again in couple hours.

Also, let me know if I should squash this last commit again :) Thank you !

NeodymiumFerBore:
Closing and reopening for CI trigger

NeodymiumFerBore:
Hello @saito-hideki  @Akasurde . I have the following error in the CI, on all Remote 2.9 : [`ERROR! couldn't resolve module/action 'community.general.filesize'. This often indicates a misspelling, missing collection, or incorrect module path.`](https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=42884&view=logs&j=dee6b80d-da0c-507c-d3ae-b401a26bd992&t=21c54440-5585-585f-94de-5cec227d4f1d&l=382). I see there is the same error on upstream CI, commit [2d3f55c](https://github.com/ansible-collections/ansible.posix/commit/2d3f55caa73559736fdd378a05dd55bba4722fcc).

Should I not worry about this? Do you have an idea why this module is not resolved on an older version of Ansible?

Also, I have some `ERROR: Timeout waiting for freebsd/12.0 instance` for all FreeBSD tests. Again, same problem on upstream CI. I cannot do much about this I guess.

Please, let me know if I have to change something in my PR. As of now I'm not sure what to do :(

Thank you!

saito-hideki:
Hi @NeodymiumFerBore 
thank you for the heads-up. I'll check that there is a problem on the CI processes.

saito-hideki:
Hi @NeodymiumFerBore 
I have fixed the CI issue to follow the requirements of community.general side.
If it is possible, can you rebase this PR with `main` branch? Once you rebased, I think the CI process will work properly.
Thank you for your patience!

NeodymiumFerBore:
Rebased PR branch to master 55bd196e82fc030346588a93ae55ccef5db201df

NeodymiumFerBore:
@saito-hideki Thank you for your action! However, @Akasurde requested [a change in the documentation part](https://github.com/ansible-collections/ansible.posix/pull/267/files#diff-9154b98384d8ac1c50a25d4cee1487d13edb55ca6955c4c656be9a60823d0a6eR77), stating that the `ephemeral state` was added in ansible.posix 1.4. Should I bump this to 1.5, or do you intend to publish this option in a 1.4 minor release?

saito-hideki:
@NeodymiumFerBore thanks! :)
Could you bump this to 1.5.0?

NeodymiumFerBore:
@saito-hideki Done! I amended previous commit to avoid adding more commits, while keeping track of the co-authored change.

NeodymiumFerBore:
Oh well, previous push succeeded 100% of the CI, now it failed due to a timeout waiting for an OS X instance. I legit changed 1 character in a string, so let's say my change is not guilty 😅

saito-hideki (approved):
LGTM :)

saito-hideki (changes requested):
Sorry, just one thing. we need to fix the version number:

[before]
This option is added in version 1.4.0.
[after]
This option is added in version 1.5.0.

plugins/modules/mount.py
```diff
@@ -68,6 +68,13 @@
       - If C(unmounted), the device will be unmounted without changing I(fstab).
       - C(present) only specifies that the device is to be configured in
         I(fstab) and does not trigger or require a mount.
+      - C(ephemeral) only specifies that the device is to be mounted, without changing
+        I(fstab). If it is already mounted, a remount will be triggered.
+        This will always return changed=True. If the mount point I(path)
+        has already a device mounted on, and its source is different than I(src),
+        the module will fail to avoid unexpected unmount or mount point override.
+        If the mount point is not present, the mount point will be created.
+        The I(fstab) is completely ignored. This option is added in version 1.4.0.
```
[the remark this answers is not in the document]
NeodymiumFerBore: Oops, stupid me did not commit the change. Sorry for that. I did as I said in [my previous comment](https://github.com/ansible-collections/ansible.posix/pull/267#issuecomment-1145675009):

> I amended previous commit to avoid adding more commits, while keeping track of the co-authored change.

saito-hideki (approved):
Approved.
Thanks! looks good to me :)

saito-hideki:
@Akasurde Finally, this PR passed CI tests without issues. I would appreciate it if you could review this PR.

NeodymiumFerBore:
Hello @saito-hideki , @Akasurde ,
Any news about this PR? Is there anything blocking it from being merged? Should I rebase it on main?

saito-hideki:
@NeodymiumFerBore sorry for the late reply. This PR has been merged.
Thank you for your contribution! :)

PR #320, opened 2022-01-29
firewalld: Add support for configuring intra zone forwarding

Ev3rt:
##### SUMMARY
Added support for configuring [intra zone forwarding](https://firewalld.org/2020/04/intra-zone-forwarding) to the ansible.posix.firewalld module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ansible.posix.firewalld

##### ADDITIONAL INFORMATION
Querying whether the forwarding option is enabled/disabled is already supported by the ansible.posix.firewalld_info module, so it would be logical to be able to configure it from the ansible.posix.firewalld module as well.

### DIFF
--- a/changelogs/fragments/320_firewalld_intra_zone_forwarding.yml
+++ b/changelogs/fragments/320_firewalld_intra_zone_forwarding.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+- firewalld - Added parameter ``forward`` to support enabling/disabling intra-zone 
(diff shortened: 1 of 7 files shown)


Ev3rt:
Interestingly these last few failing tests all fail when firewalld is disabled (offline mode). The changed setting doesn't seem to be retained. I'll have a look at the firewalld source in the next few days to hopefully understand what is happening.

PR #344, opened 2022-03-25
Support templating in ansible_user

jovial:
##### SUMMARY

Before this change it is not possible to use templating for the ansible_user e.g:

```
  ansible_user: "{{ some_variable }}"
```

The error you see looks like:

```
Unexpected remote arg: }}@10.0.0.97:/tmp/ansible.lq33g1agrally-results/\nrsync error: syntax or usage error (code 1) at main.c(1354) [sender=3.1.3]\n"
```

The module arguments look like:

```
        "module_args": {                                                                                                                                                      
            "_local_rsync_password": null,                                                                                                                                    
            "_local_rsync_path": "rsync",                                                                                                                                     
            "_substitute_controller": false,                                                                                                                                  
            "archive": false,                                                                                                                                                 
            "checksum": false,                                                                                                                                                
            "compress": true,                                                                                                                                                 
            "copy_links": false,                                                                                                                                              
            "delay_updates": true,                                                                                                                                            
            "delete": false,                                                                                                                                                  
            "dest": "/stack/tempest-artifacts",                                                                                                                               
            "dest_port": null,                                                                                                                                                
            "dirs": false,                                                                                                                                                    
            "existing_only": false,                                                                                                                                           
            "group": null,                                                                                                                                                    
            "link_dest": null,                                                                                                                                                
            "links": null,                                                                                                                                                    
            "mode": "pull",                                                                                                                                                   
            "owner": null,                                                                                                                                                    
            "partial": false,                                                                                                                                                 
            "perms": null,                                                                                                                                                    
            "private_key": null,                                                                                                                                              
            "recursive": true,                                                                                                                                                
            "rsync_opts": [],                                                                                                                                                 
            "rsync_path": null,                                                                                                                                               
            "rsync_timeout": 0,                                                                                                                                               
            "set_remote_user": true,                                                                                                                                          
            "src": "{{ kayobe_ansible_user }}@10.0.0.97:/tmp/ansible.lq33g1agrally-results/",                                                                                 
            "ssh_args": "-C -o ControlMaster=auto -o ControlPersist=60s",                                                                                                     
            "ssh_connection_multiplexing": false,                                                                                                                             
            "times": null,                                                                                                                                                    
            "verify_host": false                                                                                                                                              
        } 
```

In this case I'm using:
```
ansible_user: "{{ kayobe_ansible_user }}"
```
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

synchronize

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/action/synchronize.py
+++ b/plugins/action/synchronize.py
@@ -347,6 +347,7 @@ class ActionModule(ActionBase):
 
                 else:
                     user = task_vars.get('ansible_user') or self._play_context.remote_user
+                user = self._templar.template(user)
 
             # Private key handling
             # Use the private_key parameter if passed else use context private_key_file


plugins/action/synchronize.py
```diff
@@ -347,6 +347,7 @@ def run(self, tmp=None, task_vars=None):
 
                 else:
                     user = task_vars.get('ansible_user') or self._play_context.remote_user
+                user = self._templar.template(user)
```
bcoca, line 350: the vars were always the wrong thing to get, the correct info should be from the connection itself using  `get_option`

jovial: Seems like a good suggestion. Would we need to parse (ssh_args)[https://github.com/ansible-collections/ansible.posix/blob/7ac665a0aa169c0f3f4e0b586eb5998f481001cd/plugins/action/synchronize.py#L182-L188] to get the user? Or is that exposed as an option? Do we need to consider other connection plugins other than ssh?

maxamillion: ping @bcoca @jovial - this conversation is a bit stale :)

bcoca: all connections at this point should be relying on `get_option` some older ones might still be using `play_context.<data>` but this is incorrect as it does not always reflect changes do to modifiers in the task (mostly loops).

PR #420, opened 2023-02-23
bindep - install rsync on all EL variants

austlane:
Adds support for installing rsync for centos and related EL variants (AlmaLinux, Rocky, Oracle, etc)

rsync is an extremely common package, available in the base repos of all EL distros.

This is necessary to properly support [AWX-EE](https://github.com/ansible/awx-ee) and other community-built EL Execution Environments.

See: https://github.com/ansible/awx-ee/issues/167

### DIFF
--- a/bindep.txt
+++ b/bindep.txt
@@ -1,4 +1,4 @@
 # This is a cross-platform list tracking distribution packages needed by tests;
 # see https://docs.openstack.org/infra/bindep/ for additional information.
 
-rsync [platform:rhel-8 platform:rhel-9]
+rsync [platform:redhat]


bindep.txt
```diff
@@ -1,4 +1,4 @@
 # This is a cross-platform list tracking distribution packages needed by tests;
 # see https://docs.openstack.org/infra/bindep/ for additional information.
 
-rsync [platform:rhel-8 platform:rhel-9]
+rsync [platform:redhat]
```
austlane, line 4:
```suggestion
rsync [platform:rpm]
```
Here's an alternative solution suggested by @shanecmd
This makes the assumption that the `rsync` package exists on all other rpm-based distros.
Probably a fair assumption.

saito-hideki (approved):
Thank you for this PR.
It looks good to me :)

PR #423, opened 2023-03-09
docs: firewalld: use both `permanent` and `immediate`

jktjkt:
##### SUMMARY

docs: firewalld: use both `permanent` and `immediate`

While the docs quite clearly say what each of these options is doing, defaults matter. I'm only using Ansible every now and then (and most of the time it's for writing my CI playbooks for Zuul-CI, not for configuring system services). I took a shortcut by looking at the examples at the bottom of the doc page, and I almost wrote a loop with a dictionary for enabling both HTTP and HTTPS services and making that change immediate *and* permanent at the same time, while cursing about those ****** who wrote a crappy module which -- apparently -- has to be executed twice to do the sensible thing because everything is controlled by a single boolean. Then I discovered the `immediate` parameter, and realized that it is in fact controlled by a pair of booleans. Then I consulted my other playbook which configures a system, not a CI job, and discovered that I knew about these booleans back in 2019 when I used the firewalld module for the last time. Maybe I am the ***** after all.

Since it's perfectly OK to set both variables to true, and since people tend to want to use both in sync anyway (idempotent configurations and what not), let's use that in the examples. Examples matter.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
`ansible.posix.firewalld`

### DIFF
--- a/docs/ansible.posix.firewalld_module.rst
+++ b/docs/ansible.posix.firewalld_module.rst
@@ -415,28 +415,33 @@ Examples
       ansible.posix.firewalld:
         service: https
         permanent: yes
+        immediate: yes
         state: enabled
 
     - name: do not permit traffic in default zone on port 8081/tcp
       ansible.posix.firewalld:
         port: 8081/tcp
         permanent: yes
+        immediate: yes
         state: disabled
 
     - ansible.posix.firewalld:
         port: 161-162/udp
         permanent: yes
+        immediate: yes
         state: enabled
 
     - ansible.posix.firewalld:
         zone: dmz
         service: http
         permanent: yes
+        immediate: yes
         state: enabled
 
     - ansible.posix.firewalld:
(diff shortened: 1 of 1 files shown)


saito-hideki:
recheck

saito-hideki (changes requested):
@jktjkt thank you for this PR.
Normally, document files under docs/*.rst are generated from module code like firewalld.py by a maintainer during the release process using collection_prep.
So if it is possible, can you apply this fix to `plugins/modules/firewalld.py` instead of `docs/ansible.posix.firewalld_module.rst` ?

F.Y.I. https://github.com/ansible-network/collection_prep

maxamillion:
Changes were requested over six months ago and now there are merge conflicts, please re-open if there is still interest in contributing this patch. Thank you!

PR #335, opened 2022-03-12
allowing `synchronize` to elevate permissions when `sudo` requires password entry - implements #334

markasbach:
##### SUMMARY
This PR implements feature request #334, so that `ansible.posix.synchronize` will work with `become: yes` even if the ssh account is not set up for passwordless sudo.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`ansible.posix.synchronize`

##### ADDITIONAL INFORMATION
Steps to reproduce:
- On the target machine, set up an account with publickey authorization that is part of the `sudo` group, but make sure it has not been set up for `NOPASSWD` in the `sudoers` configuration.
- Use `ansible.posix.synchronize` with `become: yes` and use the account described above.
- The module will fail to complete because the invocation of `sudo rsync` on the remote machine brings up a password prompt and lets `rsync` fail.

With this PR, it will no longer fail.

I've been trying to follow the community guidelines for setting up this PR, but I will need some help regarding the following topics:
- I'm not sure if I can use a hardcoded `/bin/sh` for invoking `bash` within the wrapping code that I introduce in `plugins.modules.synchronize.py`. Hope this is all correct.
- Could someone cross-check I've not broken quoting / escaping on the strings that I manipulate?
- I don't know how I can add testing for this feature or actually test if my changes break anything. The repository seems to fail tests on github already.

Thanks in advance ;-)

Issue #334, opened 2022-03-12
Making `ansible.posix.synchronize` elevate permissions on accounts that need sudo password entry


markasbach:
##### SUMMARY
The current documentation of `ansible.posix.synchronize` mentions

> Currently, synchronize is limited to elevating permissions via passwordless sudo.  This is because rsync itself is connecting to the remote machine and rsync doesn't give us a way to pass sudo credentials in.

This is a substantial limitation as it requires to set up passwordless sudo (which is considered a security risk to some) to use this module with `become: yes`. It would be great to also have it work for accounts that require sudo password entry.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`ansible.posix.synchronize`

##### ADDITIONAL INFORMATION
The current limitation is due to the fact that `ansible.posix.synchronize` uses `rsync` from the target machine and therefore the typical methods to elevate permissions that are implemented in `ansible` itself cannot be used. The module already uses kind of a trick to gain root privileges on the target machine by telling `rsync` to use `--rsync-path="sudo rsync"` which works fine if the account is set up for passwordless sudo. If not, `sudo` will prompt for the password and the command fails.

If tried to find solutions for the issue and (based on a smart person's idea) found a way to pass the password safely by wrapping the `ssh` command supplied to `rsync` via `--rsh`. This is documented on https://askubuntu.com/a/1397261/874618.

I'll put up a pull request to implement this, but I might need some guidance.

tdltdc:
Could this be looked at?

This issue makes the simple act of copying more than a couple of files to privileged directories on target systems with password-protected `sudo` insufferable.

Alternatives (for example as described [here](https://stackoverflow.com/questions/27985334/why-is-copying-a-directory-with-ansible-so-slow) are not adequate:
- `ansible.builtin.copy` is easily 200x slower than `rsync` or `scp`. Copying a directory of, say, 200 small files with a total size of 1MB can take over half an hour.
- Create a temporary local tarball of the source directory, transferring and unarchiving it to the target directory with `ansible.builtin.unarchive` and then removing the temporary local tarball again "works", but is an awkward and insufficient workaround, e.g.:
  -  not idempotent
  - no integrity checks
  - always a full overwite even if the target directory is identical
  - adds unnecessary delay because the contents source directory (possibly on an HDD) first have to be read into a tarball, which then has to be read itself to be copied to the target, and then has to be read again on the target to be extracted
[...]

PR #453, opened 2023-04-25
allowing synchronize to elevate permissions when sudo requires password entry - implements #334

tdltdc:
This is a copy of #335 rebased against the current `main` branch so that it passes the CI checks.

See #335 for description and details.


tdltdc:
Never mind, let's first work with the original pull request (#335).

PR #460, opened 2023-05-13
Respawn modules to use the system python interpreter

gotmax23:
##### SUMMARY

The seboolean, selinux, firewalld, and firewalld_info modules depend on
system bindings that are only available for the default system python
interpreter. ansible-core is not packaged for the default system python
interpreter on RHEL 8 and 9. When automatic interpreter discovery does
not occur (e.g. when using implicit localhost [1]), ansible-core will
not use the system interpreter to run ansible modules and the
aforementioned modules will not work even if the bindings are installed.
The RHEL ansible-core maintainers as well as the EPEL ansible and
ansible-collection-* package maintainers (inc. me) have gotten multiple
bug reports about this. We have been telling people to fix their setup
to use the correct Python interpreter. Fortunately, ansible-core 2.11
and above have a module utility that'll respawn modules to use the
correct system interpreter.

[1] https://docs.ansible.com/ansible/latest/inventory/implicit_localhost.html

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

seboolean
selinux
firewalld
firewalld_info

### DIFF
--- a/changelogs/fragments/460-respawn.yaml
+++ b/changelogs/fragments/460-respawn.yaml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - "seboolean - respawn module to use the system python interpreter when the ``selinux`` python module is not available for ``ansible_python_interpreter``
+    (https://github.com/ansible-collections/ansible.posix/pull/460)."
--- a/changelogs/fragments/460-respawn.yaml
+++ b/changelogs/fragments/460-respawn.yaml
@@ -4,3 +4,5 @@ minor_changes:
     (https://github.com/ansible-collections/ansible.posix/pull/460)."
   - "selinux - respawn module to use the system python interpreter when the ``selinux`` python module is not available for ``ansible_python_interpreter``
     (https://github.com/ansible-collections/ansible.posix/pull/460)."
+  - "firewalld - respawn module to use the system python interpreter when the ``firewall`` python module is not available for ``ansible_python_interpreter``
+    (https://github.com/ansible-collections/ansible.posix/pull/460)."
--- a/changelogs/fragments/460-respawn.yaml
+++ b/changelogs/fragments/460-respawn.yaml
@@ -6,3 +6,5 @@ minor_changes:
     (https://github.com/ansible-collections/ansible.posix/pull/460)."
   - "firewalld - respawn module to use the system python interpreter when the ``firewall`` python module is not available for ``ansible_python_interpreter``
     (https://github.com/ansible-collections/ansible.posix/pull/460)."
+  - "firewalld_info - respawn module to use the system python interpreter when the ``firewall`` python module is not available for ``ansible_python_interpreter``
+    (https://github.com/ansible-collections/ansible.posix/pull/460)."
--- a/changelogs/fragments/460-respawn.yaml
+++ b/changelogs/fragments/460-respawn.yaml
@@ -2,3 +2,5 @@
 minor_changes:
   - "seboolean - respawn module to use the system python interpreter when the ``selinux`` python module is not available for ``ansible_python_interpreter``
     (https://github.com/ansible-collections/ansible.posix/pull/460)."
+  - "selinux - respawn module to use the system python interpreter when the ``selinux`` python module is not available for ``ansible_python_interpreter``
+    (https://github.com/ansible-collections/ansible.posix/pull/460)."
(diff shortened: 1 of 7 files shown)


gotmax23:
recheck

gotmax23:
The FreeBSD tests timed out

gotmax23:
I marked this as draft. I'd like to refactor this so that there's less duplicated code.

gotmax23:
Ready for review

maxamillion (approved):
This is amazing, thank you! 👍

maxamillion:
rebuild_merge

richm:
> ##### SUMMARY
> 
> The seboolean, selinux, firewalld, and firewalld_info modules depend on system bindings that are only available for the default system python interpreter. ansible-core is not packaged for the default system python interpreter on RHEL 8 and 9.

@gotmax23 can you explain what this means - "ansible-core is not packaged for the default system python interpreter on RHEL 8 and 9"?
Does this mean "ansible-core is packaged in such a way as to use a python module by default instead of the standard system python e.g. /usr/bin/python3.9 on RHEL 9"?
Does this mean "ansible-core does not have built in to its list of python interpreters to use for modules on managed nodes the correct python versions for RHEL 8 and RHEL 9"?

I'm wondering if this is the same problem as https://access.redhat.com/solutions/6726561:
"Issue

    When running a playbook utilizing the implicit localhost that calls a RHEL System Role on RHEL 8.6 or later with Ansible Core, the playbook fails with messages similar to:
```
TASK [redhat.rhel_system_roles.selinux : Set SELinux booleans] *****************************************************************************
...
"msg": "Failed to import the required Python library (libselinux-python) on host01's Python /usr/bin/python3.11. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"
```

If this is the problem - what we tell RHEL System Roles customers is this.

"
Root Cause

    RHEL 8.6 uses Ansible Core 2.12 to support RHEL System Roles. Ansible Core 2.12 utilizes Python 3.8 which does not contain all of the required Python modules (for example the blivet module for the storage RHEL System Role).
"

and these are the workarounds:
"

Resolution

Choose one of the options below to workaround the issue:

*    Create an inventory file that lists localhost with the ansible_connection=local option.
        For example, an inventory file with:
            localhost  ansible_connection=local
        Run ansible-playbook and specify that this inventory file should be used:
            ansible-playbook -i inventory <playbook>

*    Create an inventory file that lists localhost.
        Note that this will result in ansible-playbook connecting to the localhost over SSH with SSH key authentication, which must have previously been configured.
        For example, an inventory file with:
            localhost
        Run ansible-playbook and specify that this inventory file should be used:
            ansible-playbook -i inventory <playbook>

*    Use implicit localhost, with the ansible_python_interpreter variable set to use platform-python
        For example:
            ansible-playbook  <playbook>  -e 'ansible_python_interpreter=/usr/libexec/platform-python'
"

Unfortunately the solution in this PR is causing a problem with the way we 'vendor' in the ansible.posix modules for RHEL customers (since ansible.posix is unsupported for RHEL customers).  There is some deep failure when the python code is loaded, before the module is even executed.  I've been beating my head against the wall trying to figure out why this line (as part of the vendoring, we rewrite the module) is causing the problems:

```python
from ansible.module_utils.selinux_lsr._respawn import respawn_module, HAS_RESPAWN_UTIL
```

Also note the older legacy role form (we vendor the module into the rhel-system-roles.selinux role).

At any rate, if this is the same problem as I linked to above https://access.redhat.com/solutions/6726561 - then I think I can just remove the respawn import line, and create mock `respawn_module` and `HAS_RESPAWN_UTIL`

> When automatic interpreter discovery does not occur (e.g. when using implicit localhost [1]), ansible-core will not use the system interpreter to run ansible modules and the aforementioned modules will not work even if the bindings are installed. The RHEL ansible-core maintainers as well as the EPEL ansible and ansible-collection-* package maintainers (inc. me) have gotten multiple bug reports about this. We have been telling people to fix their setup to use the correct Python interpreter. Fortunately, ansible-core 2.11 and above have a module utility that'll respawn modules to use the correct system interpreter.
> 
> [1] https://docs.ansible.com/ansible/latest/inventory/implicit_localhost.html
> ##### ISSUE TYPE
> 
>     * Feature Pull Request
> 
> 
> ##### COMPONENT NAME
> 
> seboolean selinux firewalld firewalld_info

PR #466, opened 2023-05-21
Fix CI: replace `include:` with `include_tasks:`

felixfontein:
##### SUMMARY
`include:` is removed for ansible-core 2.16 and no longer works with current devel.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
integration tests

### DIFF
--- a/changelogs/fragments/466-tests.yml
+++ b/changelogs/fragments/466-tests.yml
@@ -1,0 +1,2 @@
+trivial:
+  - "Fix integration tests so they work with ansible-core devel / 2.16 (https://github.com/ansible-collections/ansible.posix/pull/466)."
--- a/tests/integration/targets/acl/tasks/main.yml
+++ b/tests/integration/targets/acl/tasks/main.yml
@@ -17,7 +17,7 @@
 
 - block:
 
-  - include: acl.yml
+  - include_tasks: acl.yml
     when: ansible_system == 'Linux'  # TODO enable acls mount option on FreeBSD to test it there too
 
   always:
--- a/tests/integration/targets/seboolean/tasks/main.yml
+++ b/tests/integration/targets/seboolean/tasks/main.yml
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
-- include: seboolean.yml
+- include_tasks: seboolean.yml
   when:
     - ansible_selinux is defined
     - ansible_selinux != False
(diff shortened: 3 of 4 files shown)


oraNod:
@felixfontein Hi, I've looked at the failing CI test and it looks like it's a problem with FreeBSD 12.2 trying to bootstrap from http://pkg.freebsd.org/FreeBSD:12:amd64/release_2 and getting a 404.

I think the solution is to remove a couple of targets:

https://github.com/ansible-collections/ansible.posix/blob/f48515aad3f253fd7f8a2830b42b0bd0fa6cd727/.azure-pipelines/azure-pipelines.yml#L296

https://github.com/ansible-collections/ansible.posix/blob/f48515aad3f253fd7f8a2830b42b0bd0fa6cd727/.azure-pipelines/azure-pipelines.yml#L314

I can do that in a separate PR but does that sound reasonable to you? I'm not so familiar with these tests or this repo so thought I'd ask. I can just send the PR if you're not sure either. Thanks.

felixfontein:
@oraNod I think removing them in a separate PR would be better. I'm not really familiar with this collection either, though, so no idea what the maintainers think :)

saito-hideki:
Hi @felixfontein @oraNod
Sorry if I misunderstand the situation but #466 and #467 seem deadlocked to each other.
To remove FreeBSD12.2 from CI, devel branch tests for ansible-core must pass. For that, we need to replace include with include_tasks.
Therefore, I think that the two PRs are inseparable. (I agree with you guys that it should be separated, but it seems difficult.)

felixfontein:
@saito-hideki did you try adding the gate label for one of the PRs? The (voting) Zuul tests pass in both PRs, so maybe it works?

saito-hideki (approved):
LGTM!

felixfontein:
Wow, the docker-2.9 sanity tests are still downloading a Docker image after 20 minutes...

saito-hideki:
@felixfontein I guess I misunderstood how Gate works. Thanks for the valuable advice!
Let's see how it goes :)

felixfontein:
I would hope that gate also checks AZP, but I don't think it is aware of it (because most collections that use Zuul only use Zuul and no other CI system at the same time) :)

felixfontein:
Indeed, it just ignores AZP... So I guess you have to manually pay attention to AZP status when gating...

felixfontein:
@oraNod @saito-hideki thanks for getting this in :)

saito-hideki:
@felixfontein @oraNod thank you for your help! :)

Issue #483, opened 2023-07-22
authorized_key: user option is not respected/does not work as expected


Tronde:
##### SUMMARY
After a user account was created by using the modules `ansible.builtin.group` and `ansible.builtin.user` I would like to use `ansible.posix.authorized_key` with the `user` option to configure the `authorized_keys` file of this new created user. But instead of the users's `authorized_keys` file the one of `root` is edited instead.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

authorized_key

##### ANSIBLE VERSION

```paste below
ansible [core 2.14.2]
  config file = /home/tronde/ansible/ansible.cfg
  configured module search path = ['/home/tronde/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.11/site-packages/ansible
  ansible collection location = /usr/share/ansible/collections:/home/tronde/ansible/my_collections:/home/tronde/.ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.11.2 (main, May 24 2023, 00:00:00) [GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] (/usr/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
# /home/tronde/.ansible/collections/ansible_collections
Collection    Version
------------- -------
ansible.posix 1.5.2
```

##### CONFIGURATION

```paste below
COLLECTIONS_PATHS(/home/tronde/ansible/ansible.cfg) = ['/usr/share/ansible/collections', '/home/tronde/ansible/my_collections', '/home/tronde/.ansible/collection>
CONFIG_FILE() = /home/tronde/ansible/ansible.cfg
DEFAULT_HOST_LIST(/home/tronde/ansible/ansible.cfg) = ['/home/tronde/ansible/hosts']
DEFAULT_PRIVATE_KEY_FILE(/home/tronde/ansible/ansible.cfg) = /home/jkastning/.ssh/ansible_id_rsa
DEFAULT_ROLES_PATH(/home/tronde/ansible/ansible.cfg) = ['/home/tronde/ansible', '/home/tronde/ansible/roles', '/usr/share/ansible/roles']
HOST_KEY_CHECKING(/home/tronde/ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/home/tronde/ansible/ansible.cfg) = /usr/bin/python3.9
```

[...]

PR #490, opened 2023-09-02
Docs/authorized_key: clarify that the path key should probably NOT be set

atom4git:
##### SUMMARY

Docs: Fixed unclearance in documentation connected wirh relative path

Added additional description in documentation.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

authorized_key.py

##### ADDITIONAL INFORMATION

Clarify the documentation unclearance in connected with relative path `~/.ssh/authorized_keys`
The purpose of the pull request is to eliminate ambiguities in the documentation.
In our case, when using the ~ sign, we get the user's root directory (although we explicitly specify a different username)
Here is the issue and full picture of problem which we want to fix: [LINK](https://github.com/ansible-collections/ansible.posix/issues/483)

Closese:  #483

### DIFF
--- a/plugins/modules/authorized_key.py
+++ b/plugins/modules/authorized_key.py
@@ -28,8 +28,10 @@ options:
     required: true
   path:
     description:
-      - Alternate path to the authorized_keys file.
-      - When unset, this value defaults to I(~/.ssh/authorized_keys).
+      - Alternative path to the authorized_keys file.
+      - The default value is the C(.ssh/authorized_keys) of the home of the user specified in the O(user) parameter.
+      - Most of the time, it's not necessary to set this key.
+      - Use the path to your target authorized_keys if you need to explicitly point on it.
     type: path
   manage_dir:
     description:
--- a/changelogs/fragments/490_doc_authorized_key_path.yml
+++ b/changelogs/fragments/490_doc_authorized_key_path.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - "Bugfix in the documentation regarding the path option for authorised_key(https://github.com/ansible-collections/ansible.posix/issues/483)."


goneri (changes requested):

goneri (changes requested):
Could you also please fix the commit message.

plugins/modules/authorized_key.py
```diff
@@ -28,8 +28,7 @@
     required: true
   path:
     description:
-      - Alternate path to the authorized_keys file.
```
goneri, line 31: Please key the line

plugins/modules/authorized_key.py
```diff
@@ -28,8 +28,7 @@
...
-      - When unset, this value defaults to I(~/.ssh/authorized_keys).
```
goneri, line 32: You can remove this line, but you still need to mention the default path somewhere.

plugins/modules/authorized_key.py
```diff
@@ -28,8 +28,7 @@
...
+      - It's not necessary to set this key. The default value should be enough most of the time.
```
goneri, line 31: Please change that to `Most of the time. it's not necessary to set this key. The default value should just work.`. I'm afraid the previous sentence was not clear enough.

goneri (changes requested):

plugins/modules/authorized_key.py
goneri: Thanks, you still need to mention that `I(~/.ssh/authorized_keys)` is the default value.

goneri (changes requested):

plugins/modules/authorized_key.py
```diff
@@ -28,8 +28,9 @@
     required: true
   path:
     description:
-      - Alternate path to the authorized_keys file.
-      - When unset, this value defaults to I(~/.ssh/authorized_keys).
+     - Alternative path to the authorized_keys file.
+     - Most of the time, it's not necessary to set this key.
+     - Use the path to your target authorized_keys if you need to explicitly point on it.
```
goneri, line 33: - The default value is the C(~/.ssh/authorized_keys) of the user specified in the P(user) parameter.

goneri (approved):
Thank you @atom4git for your patience and for taking the time to clarify the documentation <3.

maxamillion:
closing and re-opening to kick CI

saito-hideki (approved):
Thanks for the PR.
I rebased this PR and fixed the indentation in the document.

PR #516, opened 2023-12-04
refactor to comply with current ansible-lint guidelines

maxamillion:
##### SUMMARY

[refactor to comply with current ansible-lint guidelines](https://github.com/ansible-collections/ansible.posix/commit/280af6e372296b99e0711a4c7aa9d59b57561090)

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/dropping-ansible29.yml
+++ b/changelogs/fragments/dropping-ansible29.yml
@@ -1,0 +1,2 @@
+major_changes:
+  - "Dropping support for Ansible 2.9, ansible-core 2.14 will be minimum required version for this release"
--- a/plugins/modules/synchronize.py
+++ b/plugins/modules/synchronize.py
@@ -26,13 +26,13 @@ options:
     description:
       - Path on the source host that will be synchronized to the destination.
       - The path can be absolute or relative.
-    type: str
+    type: path
     required: true
   dest:
     description:
       - Path on the destination host that will be synchronized from the source.
       - The path can be absolute or relative.
-    type: str
+    type: path
     required: true
   dest_port:
     description:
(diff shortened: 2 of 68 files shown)


plugins/modules/synchronize.py
```diff
@@ -412,13 +429,14 @@ def main():
...
-            ssh_args=dict(type='str'),
+            _ssh_args=dict(type='str'),
+            use_ssh_args=dict(type='bool', default=False),
             ssh_connection_multiplexing=dict(type='bool', default=False),
             partial=dict(type='bool', default=False),
             verify_host=dict(type='bool', default=False),
             delay_updates=dict(type='bool', default=True),
             mode=dict(type='str', default='push', choices=['pull', 'push']),
-            link_dest=dict(type='list', elements='str'),
+            link_dest=dict(type='list', elements='path'),
```
felixfontein, line 439: Changing the type from `str` to `path` is changing behavior (since now `AnsibleModule` does certain modifications to it). Shouldn't this be mentioned in the changelog fragment?

Also the original code explicitly **only** did these transformation in special cases:
```
    # expand the paths
    if '@' not in source:
        source = os.path.expanduser(source)
    if '@' not in dest:
        dest = os.path.expanduser(dest)
```
The code now does these transformations unconditionally. I think this is a bug and should be reverted.

changelogs/fragments/dropping-ansible29.yml
```diff
@@ -0,0 +1,2 @@
+major_changes:
+  - "Dropping support for Ansible 2.9, ansible-core 2.14 will be minimum required version for this release"
```
felixfontein, line 2: So the next release will be 2.0.0 since this is a breaking change for all users of ansible-core < 2.14?

maxamillion: Looks like it needs to be, yes.

plugins/modules/synchronize.py
maxamillion: Yes! Good catch.  👍

PR #519, opened 2023-12-25
Fix Synchronize Not Sourcing Multiple Paths

Axionize:
##### SUMMARY

rsync supports sourcing files from multiple paths simply by specifying multiple paths on the command line. When the paths are folders, the same files in subsequent paths will take precedence over those in previous paths. This is often used to "merge" multiple local folders onto a remote.

For example, if this is the folder structure on the local machine and we are copying to `/tmp/dest` on the remote machine
```
tmp
├── path_a
│   ├── file1
│   └── file2
├── path_b
│   ├── file2
│   ├── file3
│   └── file4
├── path_c
│   ├── file2
│   └── file5
```
The result would be
```
tmp
├── dest
│   ├── file1 ...with data from path_a
│   ├── file2 ...from path_c
│   ├── file3.php ...path_b
│   ├── file4.php ...from path_b
│   └── file5.php ...from path_c
```

The syntax to do this in ansible is now as follows:
```
- name: Copy files from multiple folders and merge them into dest
  ansible.posix.synchronize:
    src: 
    - /tmp/path_a/
    - /tmp/path_b/
    - /tmp/path_c/
    dest: /tmp/dest/
    recursive: True
```

My primary use case for this is having a set of files stored in `host_vars`, `group_vars`, and `all` which should be copied over to the remote with files from `host_vars` superseding `group_vars` and `group_vars` superseding all.
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

synchronize

##### ADDITIONAL INFORMATION
This does not replace specifying a singular path with a string, that still works and thus backwards compatibility is preserved.

The only change this has for existing codebases on end users is in the -vvv debug output and in the invocation args of the variable returned by `register:`. Specifically `var.invocation.module_args.src` on synchronize tasks will return a list with 1 entry being the path instead of just the entry, which is unlikely to cause any breakage. If this change in output is unacceptable though, I can spend the effort to modify the modules to use strings and splitting to prevent this.

### DIFF
--- a/plugins/action/synchronize.py
+++ b/plugins/action/synchronize.py
@@ -331,6 +331,8 @@ class ActionModule(ActionBase):
         dest = _tmp_args.get('dest', None)
         if src is None or dest is None:
             return dict(failed=True, msg="synchronize requires both src and dest parameters are set")
+        if isinstance(src, str):
+            src = [src]
 
         # Determine if we need a user@ and a password
         user = None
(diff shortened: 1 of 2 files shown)


maxamillion:
Closing and re-opening to kick CI.

maxamillion:
@Axionize can you please rebase on main? I think some of the CI failures aren't valid anymore because they've been fixed in main. Thank you!

PR #524, opened 2024-01-09
Fix CI issues

felixfontein:
##### SUMMARY
- Sanity tests fail; remove problematic Shippable-specific parts of shippable.sh script.
- FreeBSD 12.4 have apparently been removed also from older versions of ansible-test.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -211,8 +211,6 @@ stages:
               test: rhel/9.1
             - name: FreeBSD 13.2
               test: freebsd/13.2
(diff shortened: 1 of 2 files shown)


maxamillion:
@felixfontein thanks for the fix, much appreciated! I was busy sorting out my local development environment nonsense we were talking about on Matrix 🙃 ... I never did find root cause but I sorted out a reasonable workaround using the creator-ee container at the suggestion of a colleague.

https://github.com/maxamillion/dotfiles/blob/main/bashrc#L193

Anyway, thanks again!

PR #544, opened 2024-05-19
Add Support for `--system` Option in Sysctl Module

### LINKED ISSUE #512: sysctl --system
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

ansible.posix.sysctl
##### ADDITIONAL INFORMATION

running a shell command of 
```
sysctl --system
```
can't be called using the module. Currently the only way to reload sysctl is to run 
```
reload: yes
```
against every sysctl module.
Being able to do a general reload would be preferable to having to run shell or cmd calls.

```yaml
- name: Reload sysctl config
  ansible.posix.sysctl:
    state: present
    reload: yes

```

MubashirUsman:
##### SUMMARY

Sysctl module does not currently support the `--system` option, which loads the configuration files from directories in the following order.
`/etc/sysctl.d/*.conf`
`/run/sysctl.d/*.conf`
`/usr/local/lib/sysctl.d/*.conf`
`/usr/lib/sysctl.d/*.conf`
`/lib/sysctl.d/*.conf`
`/etc/sysctl.conf`

I defined `SYSCTL_DIRS`, list of directories to iterate over if `system_wide` parameter is set to true. It acts same as `sysctl -p <config-file>`, and goes over all the paths defined in `SYSCTL_DIRS`.
Fixes #512 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ansible.posix.sysctl

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/sysctl.py
+++ b/plugins/modules/sysctl.py
@@ -114,12 +114,24 @@ class SysctlModule(object):
     # success or failure.
     LANG_ENV = {'LANG': 'C', 'LC_ALL': 'C', 'LC_MESSAGES': 'C'}
 
+    # We define a variable to keep all the directories to be read, equivalent to
+    # (/sbin/sysctl --system) option
+    SYSCTL_DIRS = [
+        '/etc/sysctl.d/*.conf',
+        '/run/sysctl.d/*.conf',
+        '/usr/local/lib/sysctl.d/*.conf',
+        '/usr/lib/sysctl.d/*.conf',
+        '/lib/sysctl.d/*.conf',
+        '/etc/sysctl.conf'
+    ]
+
     def __init__(self, module):
         self.module = module
         self.args = self.module.params
 
         self.sysctl_cmd = self.module.get_bin_path('sysctl', required=True)
         self.sysctl_file = self.args['sysctl_file']
+        self.system_Wide = self.args['system_Wide']
 
         self.proc_value = None  # current token value in proc fs
         self.file_value = None  # current token value in file
--- a/plugins/modules/sysctl.py
+++ b/plugins/modules/sysctl.py
@@ -101,6 +101,7 @@ import os
 import platform
 import re
 import tempfile
+import glob
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.six import string_types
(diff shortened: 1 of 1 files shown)


MaKaNu:
What is the status of this PR?

I am also at that point.

saito-hideki (commented):
Thank you for the PR! :)
As far as I know, ```sysctl```module adds or removes the sysctl parameters specified by the ```name``` parameter (which is required) to/from the configuration file specified by ```sysctl_file```, and if the reload option is set to true, it will reload the file.

saito-hideki:
@MubashirUsman 
Thank you for the update :)

> From the ubuntu man page

As mentioned in the man page you pointed out, ```--system``` option changes the basic behavior of the sysctl module, which defines sysctl parameters in a specific file and reflects the values by reloading only that file. I was concerned about that point.
So, I would like to know the specific use cases that require this function.
The typical use case would be to set several parameters in multiple files and then reload all the files with ```sysctl --system``` to apply the changes all at once.
However, ```sysct``` module applies one parameter to one file per task. Therefore I simply thought that the typical bulk application mentioned above might be better handled using the shell module or similar. I'd like to hear your thoughts on this.

MubashirUsman:
I needed to reload two files in different directories and had to use the shell module, `sysctl --system` to make the changes take effect. I was setting up the Kubernetes cluster using Kubeadm when I faced this.

MaKaNu:
> I needed to reload two files in different directories and had to use the shell module, `sysctl --system` to make the changes take effect. Using this, the `name` parameter won't be needed and can be made optional. I was setting up the Kubernetes cluster when I had to set module parameters and load `overlay` and `br_netfilter` kernal modules in different files.

Same approach here. Was setting up a kubernetes Cluster.

saito-hideki:
@MubashirUsman @MaKaNu thank you for your valuable feedback :)
This PR changes past behavior, so it cannot be backported to `stable-1`, but I agree to merge it into the main branch.

@MubashirUsman could you add some integration tests for this change and create a changelog/fragments file?
You can see more details about the changelog:

* Docs: https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to
* Examples: https://github.com/ansible-collections/ansible.posix/tree/main/changelogs/fragments

PR #546, opened 2024-06-07
Add basic ansible-lint config to fix CI; add ansible-core 2.17 to CI

felixfontein:
##### SUMMARY
ansible-lint makes the nightly CI fail (https://dev.azure.com/ansible/ansible.posix/_build/results?buildId=114105&view=logs&j=2671e6a6-f41a-533c-2720-8ffdcf8ab96f&t=5b604a49-baaa-558f-6ab2-4a2ff646af4f) due to two rules:
- `meta-runtime[unsupported-version]`: it doesn't like that the collection supports ansible-core versions that are EOL. This rule simply doesn't make any sense, and it should be disabled by default IMO.
- `fqcn[deep]`: this rule produces false positives for files in `tests/unit/plugins/action/fixtures/`.

Also adds sanity ignore file for ansible-core 2.18 (the version used by the current `devel` branch).

##### ISSUE TYPE
- Bugfix Pull Request
- Test Pull Request

##### COMPONENT NAME
ansible-lint in CI

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,0 +1,10 @@
+---
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: 2024, Ansible Project
+
+skip_list:
+  - meta-runtime[unsupported-version]  # Tis rule doesn't make any sense
+  - fqcn[deep]  # This rule produces false positives for files in tests/unit/plugins/action/fixtures/
+exclude_paths:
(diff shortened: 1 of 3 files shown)


felixfontein:
Ref: https://github.com/ansible/ansible-lint/issues/4160 (for `fqcn[deep]`)

russoz:
It looks like tests are not using ansible-core 2.17?

felixfontein:
Indeed, they don't. I'll push a commit that changes that as well.

maxamillion (approved):

felixfontein:
@russoz @maxamillion thanks a lot for reviewing and merging!

PR #547, opened 2024-06-08
Remove included RST files

felixfontein:
##### SUMMARY
The docs/ folder currently includes RST files generated from the module documentation. This is using collection_prep, which does not support modern Ansible markup. I propose to get rid of that documentation.

Also, as the comment in README.md said:

> Galaxy will eventually list the module docs within the UI, but until that is ready, you may need to either describe your plugins etc here, or point to an external docsite to cover that information.

Galaxy now *does* support that, and the list of content was already available on the Ansible docsite for the last four years.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
README
docs/


saito-hideki (approved):
Thanks for your help. This PR looks good to me!

felixfontein:
@saito-hideki thanks for reviewing and merging!

Issue #350, opened 2022-04-02
Retirement process for skippy


saito-hideki:
##### SUMMARY

Following [the duplication information](https://docs.ansible.com/ansible/latest/collections/ansible/posix/skippy_callback.html#deprecated) in the module docs, skippy is going to remove from ansible.posix. We need to make a plan and retirement schedule for [skippy](https://docs.ansible.com/ansible/latest/collections/ansible/posix/skippy_callback.html#ansible-posix-skippy-ansible-screen-output-that-ignores-skipped-status).

```
[DEPRECATED]
Removed in: major release after 2022-06-01
Why: The ‘default’ callback plugin now supports this functionality
Alternative: ‘default’ callback plugin with ‘display_skipped_hosts = no’ option
```

- [ ] TBD

##### ISSUE TYPE
- Future plan

##### COMPONENT NAME
- ansible.posix.skippy 

##### ADDITIONAL INFORMATION
- Retire date is 2022-06-01

saito-hideki:
skippy plugin will be retired on ansible.posix 2.0.0 release scheduled on Dec 5, 2024.

saito-hideki:
[2.0.0](https://github.com/ansible-collections/ansible.posix/releases/tag/2.0.0) has been released.
Skilly has been removed in 2.0.0.

PR #556, opened 2024-07-17
Remove skippy callback

Akasurde:
##### SUMMARY

Fixes: #350

##### ISSUE TYPE
- Bugfix Pull Request


saito-hideki (approved):
Thanks for the PR!
Two years have passed since skippy plugin removal was declared, so I agree with this removal as I think it is a sufficient grace period.

PR #568, opened 2024-09-17
authorized_key: Allow local path to a key

abakanovskii:
##### SUMMARY

Add option to specify an absolute path to file with SSH key(s) for `authorized_key`

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- `authorized_key`

##### ADDITIONAL INFORMATION

Before this change you would need to get key using `ansible.builtin.slurp` or something like `ansible.builtin.command: cat <file>` with `register`
I tried to keep it as simple as possible

```paste below
# Now this is possible
- name: Set authorized keys taken from path
  ansible.posix.authorized_key:
    user: charlie
    state: present
    key: /home/charlie/.ssh/id_rsa.pub
```

### DIFF
--- a/plugins/modules/authorized_key.py
+++ b/plugins/modules/authorized_key.py
@@ -24,6 +24,7 @@ options:
   key:
     description:
       - The SSH public key(s), as a string or (since Ansible 1.9) url (https://github.com/username.keys).
+      - You can also use V(file://) prefix to search remote for a file with SSH key(s).
     type: str
     required: true
   path:
@@ -95,6 +96,12 @@ EXAMPLES = r'''
     user: charlie
     state: present
     key: https://github.com/charlie.keys
+
+- name: Set authorized keys taken from path on controller node
+  ansible.posix.authorized_key:
+    user: charlie
+    state: present
+    key: file:///home/charlie/.ssh/id_rsa.pub
 
 - name: Set authorized keys taken from url using lookup
   ansible.posix.authorized_key:
@@ -223,6 +230,7 @@ from operator import itemgetter
 from ansible.module_utils._text import to_native
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.six.moves.urllib.parse import urlparse
 
 
 class keydict(dict):
@@ -556,7 +564,7 @@ def enforce_state(module, params):
     follow = params.get('follow', False)
     error_msg = "Error getting key from: %s"
 
-    # if the key is a url, request it and use it as key source
+    # if the key is a url or file, request it and use it as key source
     if key.startswith("http"):
         try:
             resp, info = fetch_url(module, key)
@@ -569,6 +577,19 @@ def enforce_state(module, params):
 
         # resp.read gives bytes on python3, convert to native string type
         key = to_native(key, errors='surrogate_or_strict')
+
+    if key.startswith("file"):
+        # if the key is an absolute path, check for existense and use it as a key source
+        key_path = urlparse(key).path
+        if not os.path.exists(key_path):
+            module.fail_json(msg="Path to a key file not found: %s" % key_path)
+        if not os.path.isfile(key_path):
+            module.fail_json(msg="Path to a key is a directory and must be a file: %s" % key_path)
+        try:
+            with open(key_path, 'r') as source_fh:
+                key = source_fh.read()
+        except OSError as e:
+            module.fail_json(msg="Failed to read key file %s : %s" % (key_path, to_native(e)))
 
     # extract individual keys into an array, skipping blank lines and comments
     new_keys = [s for s in key.splitlines() if s and not s.startswith('#')]
--- a/tests/integration/targets/authorized_key/tasks/check_path.yml
+++ b/tests/integration/targets/authorized_key/tasks/check_path.yml
@@ -1,0 +1,32 @@
+---
+- name: Create key file for test
+  ansible.builtin.copy:
+    dest: "{{ key_path }}"
+    content: "{{ rsa_key_basic }}"
+    mode: "0600"
+
+- name: Add key using path
+  ansible.posix.authorized_key:
+    user: root
+    key: file://{{ key_path }}
+    state: present
+    path: "{{ output_dir | expanduser }}/authorized_keys"
+  register: result
+
+- name: Assert that the key was added
+  ansible.builtin.assert:
+    that:
+      - result.changed == true
+
+- name: Add key using path again
+  ansible.posix.authorized_key:
+    user: root
+    key: file://{{ key_path }}
+    state: present
+    path: "{{ output_dir | expanduser }}/authorized_keys"
+  register: result
+
+- name: Assert that no changes were applied
+  ansible.builtin.assert:
+    that:
+      - result.changed == false
--- a/tests/integration/targets/authorized_key/tasks/main.yml
+++ b/tests/integration/targets/authorized_key/tasks/main.yml
@@ -31,3 +31,6 @@
 
 - name: Test for the management of comments with key
   ansible.builtin.import_tasks: comments.yml
+
+- name: Test for specifying key as a path
+  ansible.builtin.import_tasks: check_path.yml
--- a/changelogs/fragments/568_update_authorized_key.yml
+++ b/changelogs/fragments/568_update_authorized_key.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - authorized_keys - allow using absolute path to a file as a SSH key(s) source (https://github.com/ansible-collections/ansible.posix/pull/568)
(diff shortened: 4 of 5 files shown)


saito-hideki (changes requested):
@abakanovskii thank you for the PR!
I understand that the purpose of this modification is to load the public key on the target host and register it in **~/.ssh/authorized_keys** more simple steps.
I have thought about this a bit. Instead of directly specifying the path when passing the key parameter, how about the idea of adding **file://** prefix before the path, like **file:///path_to_key** style, similar to how http[s] URLs work?
Since the authorized_key module already has functionality for handling http[s], this seems like a more consistent option.

Additionally, to distinguish between handling local files on the controller and files on the target host, adding an option like **remote_src: yes/no** as in the **ansible.builtin.copy** module might make it more intuitive to specify local files on the controller using the lookup plugin.
In this case, I think the lookup plugin can still be used, so it should not be a disruptive modification.

I would like to hear your thoughts on this.

abakanovskii:
> @abakanovskii thank you for the PR! I understand that the purpose of this modification is to load the public key on the target host and register it in **~/.ssh/authorized_keys** more simple steps. I have thought about this a bit. Instead of directly specifying the path when passing the key parameter, how about the idea of adding **file://** prefix before the path, like **file:///path_to_key** style, similar to how http[s] URLs work? Since the authorized_key module already has functionality for handling http[s], this seems like a more consistent option.
> 
> Additionally, to distinguish between handling local files on the controller and files on the target host, adding an option like **remote_src: yes/no** as in the **ansible.builtin.copy** module might make it more intuitive to specify local files on the controller using the lookup plugin. In this case, I think the lookup plugin can still be used, so it should not be a disruptive modification.
> 
> I would like to hear your thoughts on this.

That seems like a good idea since we can reuse `fetch_url` but I do not understand how to switch between Controller/Managed nodes in a module since it will be played on a Managed node so what will be the purpose of `remote_src: false` with `file:///`?

tests/integration/targets/authorized_key/tasks/check_path.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: Create key file for test
+  ansible.builtin.copy:
+    dest: "{{ key_path }}"
+    content: "{{ rsa_key_basic }}"
+    mode: "0600"
+
+- name: Add key using path
+  ansible.posix.authorized_key:
+    user: root
+    key: "{{ key_path }}"
```
saito-hideki, line 11: It probably needs a prefix like ```"file://"``` with ```key~path```

tests/integration/targets/authorized_key/tasks/check_path.yml
```diff
@@ -0,0 +1,32 @@
...
+
+- name: Assert that the key was added
+  ansible.builtin.assert:
+    that:
+      - result.changed == true
+
+- name: Add key using path again
+  ansible.posix.authorized_key:
+    user: root
+    key: "{{ key_path }}"
```
saito-hideki, line 24: It probably needs a prefix like "file://" with key~path

plugins/modules/authorized_key.py
saito-hideki, line 1: I reconsidered the implementation of ```remote_src```. It seems that implementing it as an action plugin like copy module will also be necessary, so it seems more reasonable to limit it to the files on the target host, following your initial PR. I apologize for the confusion.

So, based on your initial commitment, I made the following modifications. I'd like to hear your thoughts:
```diff
diff --git a/plugins/modules/authorized_key.py b/plugins/modules/authorized_key.py
index ea58420..0f9abb5 100644
--- a/plugins/modules/authorized_key.py
+++ b/plugins/modules/authorized_key.py
@@ -24,7 +24,7 @@ options:
   key:
     description:
       - The SSH public key(s), as a string or (since Ansible 1.9) url (https://github.com/username.keys).
-      - You can also use V(file://) prefix to search localy or remote for a file with SSH key(s) depending on O(remote_src) value.
+      - You can also use V(file://) prefix to search remote for a file with SSH key(s).
     type: str
     required: true
   path:
@@ -81,13 +81,6 @@ options:
       - Follow path symlink instead of replacing it.
     type: bool
     default: false
-  remote_src:
-    description:
-      - Influence whether key needs to be transferred or already is present remotely.
-      - If V(false), it will search for src on the controller node.
-      - If V(true) it will search for src on the managed (remote) node.
-    type: bool
-    default: false
 author: Ansible Core Team
 '''
 
@@ -109,7 +102,6 @@ EXAMPLES = r'''
     user: charlie
     state: present
     key: file:///home/charlie/.ssh/id_rsa.pub
-    remote_src: true
 
 - name: Set authorized keys taken from url using lookup
   ansible.posix.authorized_key:
@@ -234,6 +226,7 @@ import tempfile
 import re
 import shlex
 from operator import itemgetter
+from urllib.parse import urlparse
 
 from ansible.module_utils._text import to_native
 from ansible.module_utils.basic import AnsibleModule
@@ -569,11 +562,10 @@ def enforce_state(module, params):
     exclusive = params.get("exclusive", False)
     comment = params.get("comment", None)
     follow = params.get('follow', False)
-    remote_src = params.get('remote_src', False)
     error_msg = "Error getting key from: %s"
 
     # if the key is a url or file, request it and use it as key source
-    if key.startswith("http") or (key.startswith("file") and remote_src):
+    if key.startswith("http"):
         try:
             resp, info = fetch_url(module, key)
             if info['status'] != 200:
@@ -586,6 +578,18 @@ def enforce_state(module, params):
         # resp.read gives bytes on python3, convert to native string type
         key = to_native(key, errors='surrogate_or_strict')
 
+    elif key.startswith("file"):
+        # if the key is an absolute path, check for existense and use it as a key source
+        key_path = urlparse(key).path
+        if not os.path.exists(key_path):
+            module.fail_json(msg="Path to a key file not found: %s" % key)
+        if not os.path.isfile(key_path):
+            module.fail_json(msg="Path to a key is a directory and must be a file: %s" % key)
+        try:
+            with open(key_path, 'r') as source_fh:
+                key = source_fh.read()
+        except OSError as e:
+            module.fail_json(msg="Failed to read key file %s : %s" % (key, to_native(e)))
     # extract individual keys into an array, skipping blank lines and comments
     new_keys = [s for s in key.splitlines() if s and not s.startswith('#')]
 
@@ -698,7 +702,6 @@ def main():
             comment=dict(type='str'),
             validate_certs=dict(type='bool', default=True),
             follow=dict(type='bool', default=False),
-            remote_src=dict(type='bool', default=False),
         ),
         supports_check_mode=True,
     )
```

```

**Review by saito-hideki — CHANGES_REQUESTED** [review]
I reconsidered the implementation of remote_src. It seems that implementing it as an action plugin like copy module will also be necessary, so it seems more reasonable to limit it to the files on the target host, following your initial PR. I apologize for the confusion.

Based on your initial PR, I made some changes to the handling of the key file. I'd like to hear your thoughts.
If there are no issues, I think we can merge this PR if you can address the points I mentioned in the review, make the corrections, and squash the commits into one.

Thanks again! :)

File: tests/integration/targets/authorized_key/tasks/check_path.yml (hunk 1 above)
Comment (abakanovskii) [reply tests/integration/targets/authorized_key/tasks/check_path.yml]: yes, I forgot about tests, thanks!

File: plugins/modules/authorized_key.py
Comment (abakanovskii) [reply plugins/modules/authorized_key.py:1]: Thank you for reviewing once again! Although we're adding new `import` now it seems more similar and intuitive than just parsing for root slash (`/`)

**Comment by abakanovskii** [comment]
there is an issue with python2.7 and urlparse module so I can either do something likethis
```python
try:
    from urllib.parse import urlparse
except ImportError:
     from urlparse import urlparse
```
or I can do what I did iin the last commit: no imports, just removing `file://` part
```
    file_prefix = "file://"
    if key.startswith(file_prefix):
        # if the key is an absolute path, check for existense and use it as a key source
        key_path = key[len(file_prefix):]
```

saito-hideki:
Hi @abakanovskii 
Thank you for the update!
 I completely forgot Python2 environment :(

```python
from ansible.module_utils.six.moves.urllib.parse import urlpars
```

How about using six ? I think it might work in both Python2 and Python3 environments.

tests/integration/targets/authorized_key/tasks/main.yml
```diff
@@ -31,3 +31,6 @@
 
 - name: Test for the management of comments with key
   ansible.builtin.import_tasks: comments.yml
+
+- name: Test for specifying key as a path
+  ansible.builtin.import_tasks: setup_steps.yml
```
saito-hideki, line 36:
```YAML
ansible.builtin.import_tasks: check_path.yml
```
I think it's `check_path.yml` :)

saito-hideki (changes requested):
Please fix the integration test to import `check_path.yml` .
The other code looks good to me :)

tests/integration/targets/authorized_key/tasks/main.yml
abakanovskii: yes, thank you!

saito-hideki (approved):
@abakanovskii 
Looks good to me! Thank you for your contribution :)

PR #569, opened 2024-09-18
mount.py no rmdir when state=='absent'

YvesMP:
##### SUMMARY
remove rmdir done after umounting when the requested state is 'absent'

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
mount.py

##### ADDITIONAL INFORMATION
In function main(), remove rmdir in case : if state == 'absent'. 

Unmounting a file system should not lead to delete anything that is revealed after unmounting.  
Also, it leads to an error if a non empty directory is present under the ex-mountpoint after umount : [Errno 39] Directory not empty . So umount is successfull but the ansible run is failed. Of course, it is solved on second run.

```paste below
# prepare for test
mkdir /var/my
mkdir /var/my/subdir
echo "some content" > /var/my/subdir/data

tree /var/my/
/var/my/
└── subdir
    └── data

# playbook used to mount
cat mount_mydata.yml
---
- name: mount my data
  hosts: localhost
  gather_facts: no
  become: yes
  tasks:
  - name: mount my data
    ansible.posix.mount:
      path: /var/my
      fstype: xfs
      state: mounted
      src: /dev/testVG/testLV
      # change with your own device

ansible-playbook mount_mydata.yml

PLAY [mount my data] **************************************************************************************************

TASK [mount my data] **************************************************************************************************
changed: [localhost]

PLAY RECAP ************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0

tree /var/my/
# gives something different
# may give no data or data depending on device content

# playbook to umount
cat unmount_mydata.yml
---
- name: unmount my data
  hosts: localhost
  gather_facts: no
  become: yes
  tasks:
  - name: unmount my data
    ansible.posix.mount:
      path: /var/my
      state: absent
      src: /dev/testVG/testLV
      # change with your own device
      backup: yes

# before code update, first run KO
ansible-playbook unmount_mydata.yml

PLAY [unmount my date] ************************************************************************************************

TASK [unmount my data] ************************************************************************************************
fatal: [localhost]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/libexec/platform-python"}, "changed": false, "msg": "Error rmdir /var/my: [Errno 39] Directory not empty: '/var/my'"}

PLAY RECAP ************************************************************************************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0

# and need a second run to be OK
ansible-playbook unmount_mydata.yml

PLAY [unmount my date] ************************************************************************************************

TASK [unmount my data] ************************************************************************************************
ok: [localhost]

PLAY RECAP ************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0


# after code update one run is enough
ansible-playbook unmount_mydata.yml

PLAY [unmount my date] ************************************************************************************************

TASK [unmount my data] ************************************************************************************************
changed: [localhost]

PLAY RECAP ************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

# back to original content
tree /var/my/
/var/my/
└── subdir
    └── data

```

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -899,11 +899,6 @@ def main():
                     module.fail_json(
                         msg="Error unmounting %s: %s" % (name, msg))
 
-            if os.path.exists(name):
-                try:
-                    os.rmdir(name)
-                except (OSError, IOError) as e:
-                    module.fail_json(msg="Error rmdir %s: %s" % (name, to_native(e)))
     elif state == 'unmounted':
         if ismount(name) or is_bind_mounted(module, linux_mounts, name):
             if not module.check_mode:


plugins/modules/mount.py
```diff
@@ -899,11 +899,6 @@ def main():
                     module.fail_json(
                         msg="Error unmounting %s: %s" % (name, msg))
 
-            if os.path.exists(name):
```
saito-hideki, line 902: This will change the behavior of ```state: absent```.
But I don't think the ```rmdir``` logic should be skipped. Because it breaks backward compatibility.
If you want to skip it, how about adding ```keep_mountpoint``` or ```preserve_mountpoint``` to keep the mountpoint to avoid ```Directory not empty``` error instead of removing this section.

If this is set as ```true```, you can skip ```rmdir``` section. Also, if the default value is set as ```false```, to keep backward compatibility, you can set it to ```false``` by default.

saito-hideki (changes requested):
Thank you for the PR! :)
I have left a comment as a change request.

YvesMP:
@saito-hideki 
As you suggested, I added the keep_mountpoint option to enable skipping the rmdir logic with state=changed.

saito-hideki (changes requested):
Hi @YvesMP 
Thanks for the update.
Your changes worked without any issues in my lab. If you address the following three things, we can merge this PR.

- [x] Please create a changelog fragment file according to the following docs:
  - docs: https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments
  - example: https://github.com/ansible-collections/ansible.posix/blob/main/changelogs/fragments/387_callback_output_header.yml
- [ ] Could you append some integration tests for keep_mountpoint?
  - https://github.com/ansible-collections/ansible.posix/tree/main/tests/integration/targets/mount/tasks
- [ ] Could you append some examples?
  - https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/mount.py#L143-L219

PR #577, opened 2024-10-11
Add ansible_ssh_password as alias for ansible_ssh_pass variable

neffs:
##### SUMMARY
ssh connection plugin reads password from variables [Link](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ssh_connection.html#parameter-password)
- ansible_password
- ansible_ssh_pass
- ansible_ssh_password
 
In synchronize ansible_ssh_password was missing, which was surprising, because it isn't documented for ansible.posix.synchronize

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible.posix.synchronize

### DIFF
--- a/plugins/action/synchronize.py
+++ b/plugins/action/synchronize.py
@@ -342,7 +342,9 @@ class ActionModule(ActionBase):
 
         # Determine if we need a user@ and a password
         user = None
-        password = task_vars.get('ansible_ssh_pass', None) or task_vars.get('ansible_password', None)
+        password = (task_vars.get('ansible_ssh_password', None)
+                    or task_vars.get('ansible_ssh_pass', None)
+                    or task_vars.get('ansible_password', None))
         if not dest_is_local:
             # Src and dest rsync "path" handling
             if boolean(_tmp_args.get('set_remote_user', 'yes'), strict=False):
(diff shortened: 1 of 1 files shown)


plugins/action/synchronize.py
```diff
@@ -342,7 +342,9 @@ def run(self, tmp=None, task_vars=None):
 
         # Determine if we need a user@ and a password
         user = None
-        password = task_vars.get('ansible_ssh_pass', None) or task_vars.get('ansible_password', None)
+        password = (task_vars.get('ansible_ssh_password', None)
+                    or task_vars.get('ansible_ssh_pass', None)
+                    or task_vars.get('ansible_password', None))
```
saito-hideki, line 347: This order is the password priority order. In this case, this logic is a first match, so how about this order based on the documents as follows: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py#L61-L64

```
          vars:
              - name: ansible_password
              - name: ansible_ssh_pass
              - name: ansible_ssh_password
```

plugins/action/synchronize.py
```diff
@@ -372,7 +374,9 @@ def run(self, tmp=None, task_vars=None):
                 src = self._process_origin(src_host, src, user)
                 dest = self._process_remote(_tmp_args, dest_host, dest, user, inv_port in localhost_ports)
 
-            password = dest_host_inventory_vars.get('ansible_ssh_pass', None) or dest_host_inventory_vars.get('ansible_password', None)
+            password = (dest_host_inventory_vars.get('ansible_ssh_password', None)
+                        or dest_host_inventory_vars.get('ansible_ssh_pass', None)
+                        or dest_host_inventory_vars.get('ansible_password', None))
```
saito-hideki, line 379: Same here'

saito-hideki (changes requested):
Thanks for your pull request.
I have left two comments as a change request.

Also, in accordance with the community guidelines, a changelog fragment file is required for all PR.
Therefore, could you add a changelog fragment file to this PR? You can see more details here: [Making your PR merge-worthy](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments)
You can write a short description for this PR in the bugfixes section.

For example: https://github.com/ansible-collections/ansible.posix/blob/main/changelogs/fragments/570_nfs4_acl.yml

PR #606, opened 2025-01-13
sysctl: Update custom sysctl file example

Pashugan:
##### SUMMARY
FreeBSD's [/boot/loader.conf](https://man.freebsd.org/cgi/man.cgi?loader.conf) may contain both `rc` and `sysctl`-like entries.

When I tried to edit this file with Ansible's `sysrc` module, I found out that it [does not support](https://docs.ansible.com/ansible/latest/collections/community/general/sysrc_module.html#notes) dot-separated keys. Searching the Internet wasn't helpful, but eventually, I realised I could modify the file with the `sysctl` module.

I think the existing example with `/tmp/test_sysctl.conf` is a bit contrived, so replacing it with `/boot/loader.conf` will make a real-world use case, and make it searchable on the Internet for others who run into the same issue as I did. It will also make it more consistent with the sysrc [examples](https://docs.ansible.com/ansible/latest/collections/community/general/sysrc_module.html#examples) that reference that file.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
`sysctl`

##### ADDITIONAL INFORMATION
n/a

### DIFF
--- a/plugins/modules/sysctl.py
+++ b/plugins/modules/sysctl.py
@@ -80,6 +80,13 @@ EXAMPLES = r'''
     sysctl_file: /tmp/test_sysctl.conf
     reload: false
 
+# Enable resource limits management in FreeBSD
+- ansible.posix.sysctl:
+    name: kern.racct.enable
+    value: '1'
+    sysctl_file: /boot/loader.conf
+    reload: false
+
 # Set ip forwarding on in /proc and verify token value with the sysctl command
 - ansible.posix.sysctl:
     name: net.ipv4.ip_forward


plugins/modules/sysctl.py
```diff
@@ -73,12 +73,13 @@
     state: absent
     sysctl_file: /etc/sysctl.conf
 
-# Set kernel.panic to 3 in /tmp/test_sysctl.conf
```
Akasurde, line 76: I would add a new example instead of modifying the existing one.

Akasurde (changes requested):

plugins/modules/sysctl.py
Pashugan: Updated. Thanks!

Akasurde:
@Pashugan Could you please rebase the branch to update the CI changes?

Akasurde (approved):

PR #682, opened 2025-09-29
[WIP] Ignore pylint errors caused by compatibility checks for six

saito-hideki:
##### SUMMARY

Ignore pylint errors caused by compatibility checks for six:

* pylint:ansible-bad-import-from

We plan to continue supporting Ansible Core 2.16 at this time.

##### ISSUE TYPE

- CI tests Request

##### COMPONENT NAME

- ansible.posix

##### ADDITIONAL INFORMATION
```paste below
None
```


saito-hideki:
@Andersson007 @felixfontein I would appreciate it if you could review this PR.
Thanks!

Andersson007 (approved):

.azure-pipelines/azure-pipelines.yml
```diff
@@ -55,10 +55,8 @@ stages:
               test: sanity
             - name: Units
               test: units
-            - name: Lint
-              test: lint
   - stage: Sanity_2_19
-    displayName: Ansible 2.19 sanitay & Units & Lint
+    displayName: Ansible 2.19 Sanitay & Units
```
felixfontein, line 59:
```suggestion
    displayName: Ansible 2.19 Sanity & Units
```

changelogs/fragments/682_update_ci_20250929.yml
```diff
@@ -0,0 +1,2 @@
+trivial:
+  - Update AZP CI matrix to ignore ansible-bad-import-from on six(https://github.com/ansible-collections/ansible.posix/pull/682).
```
felixfontein, line 2: This PR also seems to remove ansible-lint from CI.

.azure-pipelines/azure-pipelines.yml
saito-hideki: Done: Sanitay => Sanity

changelogs/fragments/682_update_ci_20250929.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
+  - Updatng AZP CI matrix to ignore ansible-bad-import-from on six(https://github.com/ansible-collections/ansible.posix/pull/682).
+  - Removing sandalone ansible-lint test it already covered by ansible-test.
```
felixfontein, line 3: ansible-test does not run ansible-lint.

A very stripped down version of ansible-lint runs during the Galaxy importer test, but even if that fails the importer tests succeeds. So basically ansible-lint is no longer run with this PR. (Assuming there is no ansible-lint run hidden somewhere else.)

If this is intended, it's OK from my side to merge this PR.

saito-hideki: Thank you for pointing that out! I seem to have completely misunderstood about ansible-test behavior.

1. I have reverted the Lint test changes and will perform ansible-lint test as before.
2. I noticed that the errors regarding ignoring `ansible-bad-import-from` in `ignore-2.20.txt` could be excluded from ansible-lint configuration. Therefore, excluded the `tests/sanity/` directory from ansible-lint testing. 

I think it's become more rational than before. Please share your opinion.

.ansible-lint
```diff
@@ -4,7 +4,8 @@
 # SPDX-FileCopyrightText: 2024, Ansible Project
 
 skip_list:
-  - meta-runtime[unsupported-version]  # Tis rule doesn't make any sense
+  - meta-runtime[unsupported-version]  # This rule doesn't make any sense
   - fqcn[deep]  # This rule produces false positives for files in tests/unit/plugins/action/fixtures/
```
felixfontein, line 8: How about simply adding
```suggestion
  - fqcn[deep]  # This rule produces false positives for files in tests/unit/plugins/action/fixtures/
  - sanity[cannot-ignore]
```

.ansible-lint
```diff
@@ -4,7 +4,8 @@
...
 exclude_paths:
   - changelogs/
+  - tests/sanity/
```
felixfontein, line 11: Instead of removing
```suggestion
```
?

saito-hideki: Thanks, I have fixed this point.

.ansible-lint
saito-hideki: Thanks! I have fixed this point.

Andersson007:
@saito-hideki @felixfontein thanks for the contribution!

Issue #685, opened 2025-11-04
ansible.posix - _AnsibleActionDone core 2.19 deprecation warning


MichalMironski:
##### SUMMARY

Deprecation Warning - everything seems to work fine, just to track it here.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ansible.posix
##### ANSIBLE VERSION

```paste below
ansible [core 2.19.3]
  config file = /home/mmironski/Workspaces/AWS-Cloud/fps-automation/config/ansible.cfg
  configured module search path = ['/home/mmironski/Workspaces/AWS-Cloud/fps-automation/config/library/modules', '/usr/share/ansible']
  ansible python module location = /home/linuxbrew/.linuxbrew/Cellar/ansible/12.1.0_1/libexec/lib/python3.13/site-packages/ansible
  ansible collection location = /home/mmironski/Workspaces/AWS-Cloud/fps-automation/config/collections
  executable location = /home/linuxbrew/.linuxbrew/bin/ansible
  python version = 3.13.9 (main, Oct 14 2025, 13:52:31) [GCC 12.3.0] (/home/linuxbrew/.linuxbrew/Cellar/ansible/12.1.0_1/libexec/bin/python)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```paste below
# /home/linuxbrew/.linuxbrew/Cellar/ansible/12.1.0_1/libexec/lib/python3.13/site-packages/ansible_collections
Collection    Version
------------- -------
ansible.posix 2.1.0
```

##### CONFIGURATION

```paste below
▶ ansible-config dump --only-changed
ANSIBLE_NOCOWS(/home/mmdir/config/ansible.cfg) = True
ANSIBLE_PIPELINING(/home/mmdir/config/ansible.cfg) = True
CACHE_PLUGIN(/home/mmdir/config/ansible.cfg) = pickle
CACHE_PLUGIN_CONNECTION(/home/mmdir/config/ansible.cfg) = ./ansible_facts
CACHE_PLUGIN_TIMEOUT(/home/mmdir/config/ansible.cfg) = 3600
COLLECTIONS_PATHS(/home/mmdir/config/ansible.cfg) = ['/home/mmdir/config/collections']
CONFIG_FILE() = /home/mmdir/config/ansible.cfg
[...]

PR #687, opened 2025-11-15
Fix _AnsibleActionDone deprecation in patch plugin

leegarrett:
##### SUMMARY
Minimal fix for the deprecation warning in the patch plugin.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
patch 

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/action/patch.py
+++ b/plugins/action/patch.py
@@ -20,7 +20,7 @@ __metaclass__ = type
 
 import os
 
-from ansible.errors import AnsibleError, AnsibleAction, _AnsibleActionDone, AnsibleActionFail
+from ansible.errors import AnsibleError, AnsibleAction, AnsibleActionFail
 from ansible.module_utils.common.text.converters import to_native
 from ansible.module_utils.parsing.convert_bool import boolean
 from ansible.plugins.action import ActionBase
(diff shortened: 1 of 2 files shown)


plugins/action/patch.py
```diff
@@ -46,7 +46,7 @@ def run(self, tmp=None, task_vars=None):
             elif remote_src:
                 # everything is remote, so we just execute the module
                 # without changing any of the module arguments
-                raise _AnsibleActionDone(result=self._execute_module(task_vars=task_vars))
+                return self._execute_module(task_vars=task_vars)
```
felixfontein: I think a more correct way to handle this would be:
```suggestion
                result.update(self._execute_module(task_vars=task_vars))
                return result
```
This is basically what the old code-path was doing.

Since `result` should be empty at this point, this is functionally equivalent to your patch, and your patch is closer to what code in ansible-core is doing nowadays. So maybe we should keep your version.

felixfontein (commented):
This likely needs a changelog fragment: https://docs.ansible.com/projects/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment

felixfontein:
recheck

felixfontein (approved):

felixfontein:
This fixes #685 btw.

Akasurde:
recheck

PR #690, opened 2025-11-26
Fix all deprecated module_utils imports before ansible-core 2.24 removal

barpavel:
##### SUMMARY

Fixes all deprecated `ansible.module_utils` imports across the entire collection that will be removed in ansible-core 2.24.

This PR comprehensively addresses deprecation warnings reported in #686 by updating import statements in 20 files to use the new recommended import paths, and removes 8 unused test utility files that contained deprecated imports.

**Deprecated imports replaced:**

| Deprecated import | Replacement |
|---|---|
| `ansible.module_utils._text` | `ansible.module_utils.common.text.converters` |
| `ansible.module_utils.common._collections_compat` | `collections.abc` |
| `ansible.module_utils.six.moves.shlex_quote` | `shlex.quote` |
| `ansible.module_utils.six.moves.reduce` | `functools.reduce` |
| `ansible.module_utils.six.moves.urllib.parse.urlparse` | `urllib.parse.urlparse` |
| `ansible.module_utils.six.string_types` | `basestring/str` (Python 2/3 compatible) |
| `ansible.module_utils.six.text_type` | `str` |
| `ansible.module_utils.six.PY3` | Removed (simplified Python 2/3 conditionals) |
| `ansible.module_utils.six.with_metaclass` | Native `metaclass=` syntax |
| `ansible.module_utils.six.iteritems` | `dict.items()` |

**Files fixed (20 files, 1 commit per file for easier review):**

- `plugins/action/patch.py`
- `plugins/action/synchronize.py`
- `plugins/callback/cgroup_perf_recap.py`
- `plugins/callback/json.py`
- `plugins/callback/jsonl.py`
- `plugins/callback/profile_roles.py`
- `plugins/callback/profile_tasks.py`
- `plugins/modules/acl.py`
- `plugins/modules/authorized_key.py`
- `plugins/modules/firewalld_info.py`
- `plugins/modules/mount.py`
- `plugins/modules/patch.py`
- `plugins/modules/rhel_rpm_ostree.py`
- `plugins/modules/rpm_ostree_upgrade.py`
- `plugins/modules/seboolean.py`
- `plugins/modules/synchronize.py`
- `plugins/modules/sysctl.py`
- `plugins/shell/csh.py`
- `plugins/shell/fish.py`
- `tests/unit/modules/system/test_mount.py`

**Files deleted (8 unused test utility files):**

[...]

PR #702, opened 2026-02-15
Replace deprecated module_utils imports before ansible-core 2.24 removal

kevinburke:
##### SUMMARY

ansible-core 2.20 deprecated `ansible.module_utils._text` and `ansible.module_utils.common._collections_compat`. Both will be removed in ansible-core 2.24.

Replace all occurrences across plugins, callbacks, and tests:

- `from ansible.module_utils._text import ...` → `from ansible.module_utils.common.text.converters import ...`

- `from ansible.module_utils.common._collections_compat import ...` → `from collections.abc import ...`

Also remove the now-unnecessary `pylint:ansible-bad-import-from` entries from the sanity ignore list for the affected files, and add a unit test that scans for these deprecated patterns to prevent regressions.

##### ISSUE TYPE

Bugfix

##### COMPONENT NAME

all

### DIFF
--- a/tests/unit/modules/conftest.py
+++ b/tests/unit/modules/conftest.py
@@ -10,8 +10,8 @@ import json
 import pytest
 
 from ansible.module_utils.six import string_types
-from ansible.module_utils._text import to_bytes
-from ansible.module_utils.common._collections_compat import MutableMapping
+from ansible.module_utils.common.text.converters import to_bytes
+from collections.abc import MutableMapping
 
 
 @pytest.fixture
--- a/tests/unit/test_deprecated_imports.py
+++ b/tests/unit/test_deprecated_imports.py
@@ -1,0 +1,69 @@
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import os
+import re
+
+from ansible_collections.ansible.posix.tests.unit.compat import unittest
+
+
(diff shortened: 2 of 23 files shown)


MatthewFletcher:
What was the reason for the failed pipelines? Not able to see the logs from the failures.

tests/unit/modules/conftest.py
```diff
@@ -10,8 +10,8 @@
 import pytest
 
 from ansible.module_utils.six import string_types
-from ansible.module_utils._text import to_bytes
-from ansible.module_utils.common._collections_compat import MutableMapping
+from ansible.module_utils.common.text.converters import to_bytes
+from collections.abc import MutableMapping
```
felixfontein, line 14: The collection still supports Python 2.7 for modules (since it supports ansible-core 2.16), so `collections.abc` cannot be used.

tests/unit/test_deprecated_imports.py
felixfontein, line 1: I would remove this test. ansible-test sanity already checks for these, and reported them (but they were ignored in tests/sanity/ignore-2.21.txt).

Akasurde:
superseded by https://github.com/ansible-collections/ansible.posix/pull/690

Akasurde:
Thanks @kevinburke

PR #709, opened 2026-04-23
feat: add 'commented' state to mount module

ollie-galbraith:
##### SUMMARY
Allow mounts to either be initialised with comments, or have existing mounts commented out (with a `force` flag)

##### COMPONENT NAME
`ansible.posix.mount`

##### ADDITIONAL INFORMATION
Using this for some automations where network backed mounts are being migrated to different machines during environment refreshes.

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -100,9 +100,12 @@ options:
       - V(absent_from_fstab) specifies that the device mount's entry will be
         removed from I(fstab). This option does not unmount it or delete the
         mountpoint.
+      - V(commented) specifies that the device mount's entry will be commented out in
+        I(fstab). This option does not unmount it or delete the mountpoint. If O(force)
+        is set to C(true), the line will be commented out even if it is currently active.
     type: str
     required: true
-    choices: [ absent, absent_from_fstab, mounted, present, unmounted, remounted, ephemeral ]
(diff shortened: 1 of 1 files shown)


plugins/modules/mount.py
```diff
@@ -100,9 +100,12 @@
       - V(absent_from_fstab) specifies that the device mount's entry will be
         removed from I(fstab). This option does not unmount it or delete the
         mountpoint.
+      - V(commented) specifies that the device mount's entry will be commented out in
+        I(fstab). This option does not unmount it or delete the mountpoint. If O(force)
+        is set to C(true), the line will be commented out even if it is currently active.
     type: str
     required: true
-    choices: [ absent, absent_from_fstab, mounted, present, unmounted, remounted, ephemeral ]
+    choices: [ absent, absent_from_fstab, commented, mounted, present, unmounted, remounted, ephemeral ]
```
Akasurde, line 108: I think, we can use `lineinfile` module to comment or edit mount file.

Akasurde (changes requested):

plugins/modules/mount.py
ollie-galbraith: `lineinfile` will work to comment a line for the first instance, however on every subsequent run of the playbook, `mount` wouldn't match on a commented line, so it will add a new line with the same content. Then `lineinfile` matches again and re-comments, and repeat.

PR #742, opened 2026-05-20
mount: restore fstab file if umount fails

### LINKED ISSUE #605: state=absent failure to unmount does not revert fstab changes
##### SUMMARY

When using `state: absent` to unmount a volume and remove it from `/etc/fstab/`, the module removes the fstab entry first and then unmounts the volume.

In the case of an error while unmounting (eg. `Device busy`), the fstab is not restored which puts the host in an improper state where a replay of the task/playbook will not unmount the volume since it's no longer present in `/etc/fstab`.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

mount

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.11]
  config file = None
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/.venvs/ansible-9.10.0/lib/python3.12/site-packages/ansible
  ansible collection location = ~/.ansible/collections:/usr/share/ansible/collections
  executable location = ~/.venvs/ansible-9.10.0/bin/ansible
  python version = 3.12.6 (main, Sep 17 2024, 13:46:34) [GCC 13.2.1 20231205 (Red Hat 13.2.1-6)] (~/.venvs/ansible-9.10.0/bin/python3.12)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection    Version
------------- -------
ansible.posix 1.5.4
```

##### CONFIGURATION

```paste below
ANSIBLE_FORCE_COLOR(REDACTED/ansible.cfg) = True
CONFIG_FILE() = ./ansible.cfg
DEFAULT_HASH_BEHAVIOUR(REDACTED/ansible.cfg) = merge
DEFAULT_STDOUT_CALLBACK(REDACTED/ansible.cfg) = debug
HOST_KEY_CHECKING(REDACTED/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Linux target hosts

##### STEPS TO REPRODUCE

1. Mount a volume somewhere on the target host and add it to /etc/fstab
2. Do something to force the device in a busy state
3. Launch the playbook unmounting the volume
4. Notice the error
5. Launch the playbook a second time
6. Notice it now finishes without an error
7. Get the device out of its busy state
8. Launch the playbook a third time
9. Notice it still finishes without an error
10. The device is removed from fstab but still mounted

```shell
# Step 1:
## On target host as root
VOLFILE=/path/to/volumefile # file to use as volume for this test
FSTYPE=ext2 # change with your choice of filesystem
DEVICE=$(losetup -f)
MOUNTPOINT=/mnt # set to an available mountpoint on host
## Setup our test volume
dd if=/dev/zero bs=1M count=100 of="${VOLFILE}"
losetup "${DEVICE}" "${VOLFILE}"
mkfs."${FSTYPE}" "${DEVICE}"
## Add it to /etc/fstab
echo -e "${DEVICE}\t${MOUNTPOINT}\t${FSTYPE}\tdefaults\t0 0" >> /etc/fstab
## Mount it (and validate fstab)
mount -a
# Step 2:
## Make device busy
cd "${MOUNTPOINT}"
## keep this shell open until step 7
sleep 3600
```

```yaml
---
# Change values according to your setup
- name: mount cmve
  host: target_host
  become: true
  tasks:
    - name: Unmount or test volume
      ansible.posix.mount:
        path: /mnt
        state: absent
```

##### EXPECTED RESULTS

On failure, the entry in fstab should be preserved.

This could be accomplished either by adding it back after a failure or unmounting before editing fstab.

In the previously described test case, the first AND second run should fail and only the last should succeed, leaving us with both the fstab edited and the volume uunmounted.

##### ACTUAL RESULTS

After the first failure, `/etc/fstab` doesn't contain the volume anymore even though it's still mounted.

On the second run, no action is attempted since no matching entry is present in fstab. The same goes for the third run.

```paste below
# Not relevant AFAIK
```

Akasurde:
##### SUMMARY

* Refactored code

Fixes: #605

Signed-off-by: Abhijeet Kasurde <Akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/mount_restore_fstab.yml
plugins/modules/mount.py

### DIFF
--- a/changelogs/fragments/mount_restore_fstab.yml
+++ b/changelogs/fragments/mount_restore_fstab.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+    - mount - restore fstab if umount fails (https://github.com/ansible-collections/ansible.posix/issues/605).
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -801,20 +801,24 @@ def main():
             dict(
                 opts='-',
                 passno='-',
-                fstab=module.params.get('fstab', '/etc/vfstab'),
+                fstab=module.params['fstab'],
                 boot='yes' if module.params['boot'] else 'no',
             )
         )
+        if args['fstab'] is None:
+            args['fstab'] = '/etc/vfstab'
     else:
         args.update(
             dict(
                 opts='defaults',
                 dump='0',
                 passno='0',
-                fstab=module.params.get('fstab', '/etc/fstab'),
+                fstab=module.params['fstab'],
                 boot='yes',
             )
         )
+        if args['fstab'] is None:
+            args['fstab'] = '/etc/fstab'
 
         # FreeBSD doesn't have any 'default' so set 'rw' instead
         if PLATFORM == 'freebsd':
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -229,12 +229,14 @@ from ansible.module_utils.common.text.converters import to_bytes, to_native
 from ansible.module_utils.parsing.convert_bool import boolean
 
 
+PLATFORM = platform.system().lower()
+
+
 def write_fstab(module, lines, path):
 
+    backup_file = ""
     if module.params['backup']:
         backup_file = module.backup_local(path)
-    else:
-        backup_file = ""
 
     fs_w = open(path, 'w')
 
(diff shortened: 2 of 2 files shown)


saito-hideki (approved):
LGTM!
Thanks for the PR :)

Issue #595, opened 2024-11-21
If a comment is appended to a fstab entry state present creates a double-entry .


vbotka:
##### SUMMARY

Given the /tmp/fstab for testing

```bash
shell> cat /tmp/fstab 
/dev/sda /mnt3 ntfs3 defaults 0 0 # mount sda to /mnt3
```

the module ansible.posix.mount

```yaml
    - ansible.posix.mount:
        fstab: /tmp/fstab
        src: /dev/sda
        path: /mnt3
        fstype: ntfs3
        state: present
```

creates a double-entry

```bash
shell> cat /tmp/fstab 
/dev/sda /mnt3 ntfs3 defaults 0 0 # mount sda to /mnt3
/dev/sda /mnt3 ntfs3 defaults 0 0
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ansible.posix.mount

##### ANSIBLE VERSION

```ini
shell> ansible --version
ansible [core 2.18.0]
  config file = /export/scratch/sandbox/tmp1/test-101/ansible.cfg
  configured module search path = ['/home/vlado/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/vlado/env/lib/python3.12/site-packages/ansible
  ansible collection location = /home/vlado/env/lib/python3.12/site-packages/ansible_collections
  executable location = /home/vlado/env/bin/ansible
  python version = 3.12.3 (main, Sep 11 2024, 14:17:37) [GCC 13.2.0] (/home/vlado/env/bin/python3)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```bash
shell> ansible-galaxy collection list ansible.posix

# /home/vlado/env/lib/python3.12/site-packages/ansible_collections
Collection    Version
------------- -------
ansible.posix 1.6.2
```

##### CONFIGURATION

```
ANSIBLE_PIPELINING(/export/scratch/sandbox/tmp1/test-101/ansible.cfg) = True
CONFIG_FILE() = /export/scratch/sandbox/tmp1/test-101/ansible.cfg
DEFAULT_GATHERING(/export/scratch/sandbox/tmp1/test-101/ansible.cfg) = explicit
DEFAULT_HOST_LIST(/export/scratch/sandbox/tmp1/test-101/ansible.cfg) = ['/scratch/sandbox/tmp1/test-101/hosts']
[...]

PR #744, opened 2026-05-22
mount: Update line splitting logic

### LINKED ISSUE #743: mount: Incorrect comment handling breaks device specs with '#' (SSHFS, CIFS)
##### SUMMARY
The fix merged in #595 (commit 8e34860) incorrectly strips comments from fstab lines using `line.split('#')[0].split()`, which breaks valid device specifications containing `#` characters. This affects SSHFS mounts (`sshfs#user@host:path`), CIFS mounts, and any filesystem where `#` is part of the device identifier.

**Important: Issue #595 was based on a misunderstanding of the fstab format.** End-of-line comments (text after valid fstab entries on the same line) are NOT supported in fstab according to libmount and the fstab specification. What appeared to "work" in #595 was not because `#` was being interpreted as a comment marker, but simply because libmount parses exactly 6 fields and silently ignores anything beyond field 6. The text after the 6th field is ignored regardless of whether it starts with `#` or not.

The current implementation does not match how libmount (util-linux) actually parses `/etc/fstab`. According to libmount source code, only lines beginning with `#` (after leading whitespace) are treated as comments. The `#` character appearing within field values must be preserved.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
mount

##### ANSIBLE VERSION

```
shell> ansible --version
ansible [core 2.20.6]
  config file = None
  configured module search path = ['/home/flavio/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/flavio/.micromamba/lib/python3.14/site-packages/ansible
  ansible collection location = /home/flavio/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/flavio/.micromamba/bin/ansible
  python version = 3.14.4 | packaged by conda-forge | (main, Apr  8 2026, 01:59:35) [GCC 14.3.0] (/home/flavio/.micromamba/bin/python3.14)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION
```
shell> ansible-galaxy collection list ansible.posix

# /home/flavio/.ansible/collections/ansible_collections
Collection    Version
------------- -------
ansible.posix 2.2.0
```

##### CONFIGURATION
```
shell> ansible-config dump --only-changed
CONFIG_FILE() = None
EDITOR(env: EDITOR) = /usr/bin/vi
PAGER(env: PAGER) = /usr/bin/most

GALAXY_SERVERS:
```

##### OS / ENVIRONMENT
```
shell> uname -srvmpio
Linux 6.0.11 #1 SMP PREEMPT_DYNAMIC Sun Dec 4 12:34:56 CET 2022 x86_64 12th Gen Intel(R) Core(TM) i7-1265U GenuineIntel GNU/Linux
```

##### STEPS TO REPRODUCE
```yaml
- hosts: localhost
  tasks:
    - name: Add an entry to `/etc/fstab`
      ansible.posix.mount:
        src: "sshfs#USER@HOST:/home/USER"
        path: "/home/USER/remote_home"
        fstype: fuse3
        opts: "allow_root,PasswordAuthentication=no,identityfile=/home/USER/.ssh/id_ed25519,_netdev,setuid=USER"
        state: present
```

##### EXPECTED RESULTS
The task should be idempotent. This means, the second time it is run it should report `changed=false`.

##### ACTUAL RESULTS
However, a second entry is created in `/etc/fstab/`.

```paste below
shell> cat /etc/fstab | grep sshfs
```

##### ROOT CAUSE
The code in `plugins/modules/mount.py` (line ~306 after commit 8e34860) splits on `#`:
```python
fields = line.split('#')[0].split()
```

When reading the existing entry `sshfs#USER@HOST:/home/USER`, this becomes just `sshfs`, which doesn't match the requested device specification. The module then adds a second entry with the incorrect device spec.

##### PROPOSED FIX
Replace with logic matching libmount's behavior:
```python
fields = line.split()[:6]  # libmount parses max 6 fields
```

##### REFERENCES
- Original issue: #595 (based on misunderstanding of fstab format)
- Problematic commit: 8e34860d6ccda0cd9ebb5a86e2ba365e4b94767d
- libmount source: https://github.com/util-linux/util-linux/blob/master/libmount/src/tab_parse.c
  - `is_comment_line()`: Only checks for `#` at start of line
  - `mnt_parse_table_line()`: Parses exactly 6 fields, ignores the rest
- fstab(5) man pages (OpenBSD, Linux): "Lines beginning with '#' are comments"

calvofl0:
##### SUMMARY
Fix incorrect fstab comment handling that breaks device specifications containing `#` characters (e.g., SSHFS `sshfs#user@host:path`).

The fix in commit 8e34860 for issue #595 uses `line.split('#')[0]` which incorrectly treats any `#` in a line as starting a comment. This breaks valid fstab entries where `#` is part of the device specification.

**Important context:** Issue #595 was based on a misunderstanding. End-of-line comments are NOT supported in fstab - libmount only treats `#` at the start of a line as a comment. What appeared to "work" in #595 was simply that libmount parses exactly 6 fields and ignores anything beyond field 6, regardless of whether it starts with `#`.

This PR changes the parsing to match libmount's actual behavior:
- Only treat `#` at line start (after whitespace) as a comment
- Parse maximum 6 fields and ignore extras
- Preserve `#` characters within field values

Fixes #743 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
mount

##### ADDITIONAL INFORMATION
**Before (broken):**
```python
fields = line.split('#')[0].split()
```

**After (correct):**
```python
stripped = line.strip()
if stripped.startswith('#') or not stripped:
    to_write.append(line)
    continue
fields = stripped.split()[:6]
```

**Impact:** This fixes SSHFS mounts, CIFS mounts with `#` in paths, and any filesystem using `#` in device specifications.

**Testing:**
```yaml
- name: Test SSHFS mount is idempotent
  ansible.posix.mount:
    src: "sshfs#user@host:/path"
    path: /mnt/remote
    fstype: fuse3
    opts: allow_root,_netdev
    state: present
```

First run: creates entry correctly
Second run: `changed=false` (idempotent)

**References:**
- libmount source: https://github.com/util-linux/util-linux/blob/master/libmount/src/tab_parse.c
  - `is_comment_line()`: Only checks for `#` at start of line
  - `mnt_parse_table_line()`: Parses exactly 6 fields
- fstab(5) man pages: "Lines beginning with '#' are comments"

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -302,7 +302,7 @@ def _set_mount_save_old(module, args):
 
             continue
 
-        fields = line.split('#')[0].split()
+        fields = line.split()[:6]
 
         # Check if we got a valid line for splitting
         # (on Linux the 5th and the 6th field is optional)
--- a/changelogs/fragments/743-mount-fix-hash-in-device-specs.yml
+++ b/changelogs/fragments/743-mount-fix-hash-in-device-specs.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - mount - fixed incorrect comment handling introduced in commit 8e34860 that broke device specifications containing ``#`` characters (e.g., SSHFS ``sshfs#user@host:path``) by incorrectly stripping everything after the first ``#``. The module now matches libmount behavior by only treating ``#`` at the start of a line as a comment (https://github.com/ansible-collections/ansible.posix/issues/743).


plugins/modules/mount.py
```diff
@@ -302,7 +302,7 @@ def _set_mount_save_old(module, args):
 
             continue
 
-        fields = line.split('#')[0].split()
+        fields = line.split()[:6]
```
Akasurde, line 305: Could you please add an integration test for this change? Thanks

Akasurde (changes requested):

plugins/modules/mount.py
saito-hideki, line 305: My concern is that splitting at [:6] may cause the following check in SunOS to fail.

- https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/mount.py#L310

We also need to consider operating systems other than Linux.

saito-hideki (changes requested):
Thanks for the PR!
My concern is that splitting at [:6] may cause the following check in SunOS to fail.

- https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/mount.py#L310

We also need to consider operating systems other than Linux.

calvofl0:
Why not simply reverting commit 8e34860? Comments at the end of the line (in `/etc/fstab`) are simply not supported in Linux, even if they accidentally work, and not supported in SunOS either, so that specific commit is effectively introducing a bug while solving a non-problem. Why should `ansible.posix` handle end-of-line comments?

Alternatively, if you really want, we would need a platform-dependent approach:

```
if platform.system() == 'SunOS':
    fields = line.split()[:7]
else:
    fields = line.split()[:6]
```

PR #760, opened 2026-06-16
authorized_key: Fixes the symlink handling issue reported in CVE-2026-11837

### LINKED ISSUE #759: ansible.posix authorized_key: local privilege escalation via symlink-following chown (CVE-2026-11837)
I'm filling this as the https://bugzilla.redhat.com/show_bug.cgi?id=2487424 mentions the CVE-2026-11837 assgnment and is public but am not aware of an already reported problem. The Red Hat Bugzilla entry contains:

> A local privilege escalation vulnerability was found in the ansible.posix authorized_key module. The keyfile() function uses os.chown() (not os.lchown()) and plain open() (no O_NOFOLLOW) when managing a user's ~/.ssh directory and authorized_keys file. An unprivileged local user can pre-stage symlinks in their ~/.ssh directory. When an operator runs the authorized_key task as root, the module follows the symlinks and changes ownership of arbitrary files/directories to the unprivileged user, enabling root escalation.
>
> This is a sibling of [CVE-2024-9902](https://access.redhat.com/security/cve/CVE-2024-9902) (ansible-core user module), which addressed the same symlink-following class in generate_ssh_key. The authorized_key module in the separate ansible.posix collection was not covered by that fix.

Is this issue on your radar?

saito-hideki:
##### SUMMARY

This PR fixes the code to properly handle symlinks for `sshdir` and `keysfile`.

- Addressed CVE-2026-11837
- Fixed #759 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- ansible.posix.authorized_key

##### ADDITIONAL INFORMATION
```paste below
None
```

### DIFF
--- a/plugins/modules/authorized_key.py
+++ b/plugins/modules/authorized_key.py
@@ -307,6 +307,17 @@ class keydict(dict):
         return [item[1] for item in self.items()]
 
 
+def _safe_open_write(module, path, follow):
+    flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+    if not follow and hasattr(os, 'O_NOFOLLOW'):
+        flags |= os.O_NOFOLLOW
+    try:
+        fd = os.open(path, flags, int('0600', 8))
+    except OSError as e:
+        module.fail_json(msg="File open failed %s : %s" % (path, to_native(e)))
+    return fd
+
+
 def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False):
     """
     Calculate name of authorized keys file, optionally creating the
@@ -359,7 +370,7 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
                 module.fail_json(msg="Failed to create directory %s : %s" % (sshdir, to_native(e)))
             if module.selinux_enabled():
                 module.set_default_selinux_context(sshdir, False)
-        os.chown(sshdir, uid, gid)
+        os.chown(sshdir, uid, gid, follow_symlinks=follow)
         os.chmod(sshdir, int('0700', 8))
 
     if not os.path.exists(keysfile):
@@ -367,16 +378,13 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
         if not os.path.exists(basedir):
             os.makedirs(basedir)
 
-        f = None
-        try:
-            f = open(keysfile, "w")  # touches file so we can set ownership and perms
-        finally:
-            f.close()
+        fd = _safe_open_write(module, keysfile, follow)
+        os.close(fd)
         if module.selinux_enabled():
             module.set_default_selinux_context(keysfile, False)
 
     try:
-        os.chown(keysfile, uid, gid)
+        os.chown(keysfile, uid, gid, follow_symlinks=follow)
         os.chmod(keysfile, int('0600', 8))
     except OSError:
         pass
@@ -607,7 +615,7 @@ def enforce_state(module, params):
 
     # check current state -- just get the filename, don't create file
     do_write = False
-    params["keyfile"] = keyfile(module, user, do_write, path, manage_dir)
+    params["keyfile"] = keyfile(module, user, do_write, path, manage_dir, follow)
     existing_content = readfile(module, params["keyfile"])
     existing_keys = parsekeys(module, existing_content)
 
@@ -696,6 +704,11 @@ def enforce_state(module, params):
 
         if not module.check_mode:
             writefile(module, filename, new_content)
+            user_entry = pwd.getpwnam(user)
+            uid = user_entry.pw_uid
+            gid = user_entry.pw_gid
+            os.chown(filename, uid, gid, follow_symlinks=follow)
+            os.chmod(filename, int('0600', 8))
         params['changed'] = True
 
     return params
--- a/tests/integration/targets/authorized_key/tasks/check_symlink.yml
+++ b/tests/integration/targets/authorized_key/tasks/check_symlink.yml
@@ -1,0 +1,22 @@
+---
+#
+# Check: keysfile is symlink and follow=false
+#
+- name: Try to add key with keysfile as symlink
+  ansible.posix.authorized_key:
+    user: testuser
+    key: "{{ rsa_key_basic }}"
+    state: present
+    manage_dir: false
+    follow: false
+
+- name: Assert target file ownership unchanged
+  ansible.builtin.stat:
+    path: /tmp/symlink_test_target/target_file
+  register: target_file_stat
+
+- name: Verify target file is still owned by root
+  ansible.builtin.assert:
+    that:
+      - target_file_stat.stat.uid == 0
+...
--- a/changelogs/fragments/759_cve_2026_11837.yml
+++ b/changelogs/fragments/759_cve_2026_11837.yml
@@ -1,0 +1,2 @@
+security_fixes:
+  - authorized_key - fix local privilege escalation via symlink-following when running as root (https://github.com/ansible-collections/ansible.posix/issues/759).
--- a/tests/integration/targets/authorized_key/tasks/check_symlink_cleanup.yml
+++ b/tests/integration/targets/authorized_key/tasks/check_symlink_cleanup.yml
@@ -1,0 +1,12 @@
+---
+- name: Remove testuser
+  ansible.builtin.user:
+    name: testuser
+    state: absent
+    remove: true
+
+- name: Remove symlink target directory
+  ansible.builtin.file:
+    path: /tmp/symlink_test_target
+    state: absent
+...
--- a/tests/integration/targets/authorized_key/tasks/check_symlink_setup.yml
+++ b/tests/integration/targets/authorized_key/tasks/check_symlink_setup.yml
@@ -1,0 +1,38 @@
+---
+- name: Create testuser for symlink tests
+  ansible.builtin.user:
+    name: testuser
+    create_home: true
+  register: testuser_result
+
+- name: Create symlink target directory
+  ansible.builtin.file:
+    path: /tmp/symlink_test_target
+    state: directory
+    owner: root
+    mode: '0700'
+
+- name: Create a target file owned by root
+  ansible.builtin.copy:
+    dest: /tmp/symlink_test_target/target_file
+    content: "sensitive data"
+    owner: root
+    mode: '0600'
+
+- name: Remove .ssh directory if exists
+  ansible.builtin.file:
+    path: "{{ testuser_result.home }}/.ssh"
+    state: absent
+
+- name: Create .ssh directory
+  ansible.builtin.file:
+    path: "{{ testuser_result.home }}/.ssh"
+    state: directory
+    mode: '0700'
+
+- name: Create symlink from authorized_keys to target file
+  ansible.builtin.file:
+    src: /tmp/symlink_test_target/target_file
+    dest: "{{ testuser_result.home }}/.ssh/authorized_keys"
+    state: link
+...
--- a/tests/integration/targets/authorized_key/tasks/main.yml
+++ b/tests/integration/targets/authorized_key/tasks/main.yml
@@ -17,7 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
-- name: Setup testing environment
+- name: Setup for testing environment
   ansible.builtin.import_tasks: setup_steps.yml
 
 - name: Test for multiple keys handling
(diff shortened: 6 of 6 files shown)


saito-hideki:
Hi @Akasurde @Andersson007 @bcoca 
I would appreciate it if you could review this PR.
If there are no issues, we plan to backport it to stable-2 and release version 2.2.1.
Please let me know your thoughts.

plugins/modules/authorized_key.py
```diff
@@ -307,6 +309,16 @@ def values(self):
         return [item[1] for item in self.items()]
 
 
+def _assert_symlink(module, path):
+    """
+    Addressed CVE-2026-11837.
+    If path is symlink, module should be failed with refusing message.
+    """
```
Akasurde: Drop this description.

plugins/modules/authorized_key.py
```diff
@@ -307,6 +309,16 @@ def values(self):
         return [item[1] for item in self.items()]
 
 
+def _assert_symlink(module, path):
```
Akasurde: accept follow parameter and reject only if follow is false and path is link. 

```Python
if not follow and os.path.islink(path):
    module.fail_json(
        msg="Refusing to operate on symlink: %s. Please use follow=true if this is intentional." % path)
    )

File: plugins/modules/authorized_key.py [hunk 3]
```diff
@@ -345,12 +357,19 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
     if follow:
         keysfile = os.path.realpath(keysfile)
 
+    # Reject sshdir symlinks when follow=false (CVE-2026-11837).
```
Akasurde: drop this comment.

plugins/modules/authorized_key.py
```diff
@@ -359,27 +378,50 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
                 module.fail_json(msg="Failed to create directory %s : %s" % (sshdir, to_native(e)))
             if module.selinux_enabled():
                 module.set_default_selinux_context(sshdir, False)
-        os.chown(sshdir, uid, gid)
-        os.chmod(sshdir, int('0700', 8))
+            fd = os.open(sshdir, os.O_RDONLY)
+        elif follow:
+            # If sshdir exists and follow=true, open it without O_NOFOLLOW.
+            fd = os.open(sshdir, os.O_RDONLY)
```
Akasurde: instead of branching on follow here, we can have another function to add `O_NOFOLLOW`, something like

```python
def _safe_open_write(module, path, follow):
    flags = os.O_WRONLY | os.O_CREAT
    if not follow and hasattr(os, 'O_NOFOLLOW'):
        flags |= os.O_NOFOLLOW
    try:
        fd = os.open(path, flags, int('0600', 8
    except OSError as e:
        module.fail_json(msg="Refusing to operate on symlink: %s : %s" % (sshdir, to_native(e)))
    return os.fdopen(fd, 'w')
```

Akasurde (dismissed):

plugins/modules/authorized_key.py
saito-hideki: done

plugins/modules/authorized_key.py
saito-hideki: done

plugins/modules/authorized_key.py
saito-hideki: I'll implement _safe_open_write() and _safe_open_read() and use them. Thanks!

plugins/modules/authorized_key.py
saito-hideki: I will fix _assert_symlink().

saito-hideki:
@Akasurde Thank you for reviewing this PR!
I’ve made the updates based on your feedback. Could you please take another look when you have a moment?

plugins/modules/authorized_key.py
saito-hideki: done

plugins/modules/authorized_key.py
saito-hideki: done

plugins/modules/authorized_key.py
```diff
@@ -79,6 +79,8 @@
   follow:
     description:
       - Follow path symlink instead of replacing it.
+      - When set to V(false), the module will refuse to operate if the authorized keys file or its parent directory is a symlink.
```
bcoca: this is a backwards incompatible change and contradicts the first line

plugins/modules/authorized_key.py
```diff
@@ -345,12 +375,17 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
...
+
     if not write or module.check_mode:
         return keysfile
 
     uid = user_entry.pw_uid
     gid = user_entry.pw_gid
 
+    # Ensure sshdir exists with correct ownership and permissions.
+    # Use file descriptor based fchown/fchmod to prevent symlink races
+    # where an attacker swaps sshdir for a symlink between check and operation.
```
bcoca: this does not ensure it exists with correct ownership/permissions, it creates it with correct ownership/permissions but does not verify if existing one has them

plugins/modules/authorized_key.py
```diff
@@ -359,27 +394,34 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
                 module.fail_json(msg="Failed to create directory %s : %s" % (sshdir, to_native(e)))
             if module.selinux_enabled():
                 module.set_default_selinux_context(sshdir, False)
-        os.chown(sshdir, uid, gid)
-        os.chmod(sshdir, int('0700', 8))
+            fd = os.open(sshdir, os.O_RDONLY)
```
bcoca: these changes are not required as you can simply add `follow_symlinks=follow` to handle the issues with chmod/chown or just avoid them when the target is a symlnk + follow

tests/integration/targets/authorized_key/tasks/check_symlink.yml
```diff
@@ -0,0 +1,181 @@
...
+#
+- name: Remove .ssh directory if exists
+  ansible.builtin.file:
+    path: "{{ testuser_result.home }}/.ssh"
+    state: absent
+
+- name: Create symlink from .ssh to external directory
+  ansible.builtin.file:
+    src: /tmp/symlink_test_target
+    dest: "{{ testuser_result.home }}/.ssh"
```
bcoca: the CVE is referring to symlinks inside the .ssh dir, not to the ssh dir being itself a symlink

plugins/modules/authorized_key.py
```diff
@@ -345,12 +375,17 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
     if follow:
         keysfile = os.path.realpath(keysfile)
 
+    _assert_symlink(module, sshdir, follow)
```
bcoca: if we are going to check the containing dir, it being a symlink would not be the only thing i would check, i would also add proper permissions, showing a warning or error, but then we would still need a way to allow the operation to proceed by adding something like a `force` option.

bcoca (changes requested):

saito-hideki:
@bcoca Thanks for reviewing this PR.
I think I understand your points. The issue is that the scope of the original changes is too broad and includes backward-incompatible fixes.
Based on your and @Akasurde's reviews, I'm going to revise this PR. Once the revisions are complete, I will request another review.
Please let me know if my understanding is incorrect.

Thanks!

plugins/modules/authorized_key.py
saito-hideki: Removed the documentation change. The`follow` parameter description is kept as-is.

plugins/modules/authorized_key.py
saito-hideki: Removed the misleading comment.

plugins/modules/authorized_key.py
saito-hideki: Replaced the approach with follow_symlinks=follow on os.chown() and os.chmod().

tests/integration/targets/authorized_key/tasks/check_symlink.yml
saito-hideki: Removed the sshdir symlink tests. The test now only covers the case where authorized_keys itself is a symlink.

plugins/modules/authorized_key.py
saito-hideki: Removed the sshdir symlink check to keep the fix scoped to the CVE.

saito-hideki:
recheck

saito-hideki:
Hi @Akasurde @bcoca,
I've revised it based on your feedback. Could you please take another look?

Thanks!
Hideki

saito-hideki:
recheck

gundalow:
@Akasurde @bcoca could you please re-review.
Thanks

plugins/modules/authorized_key.py
```diff
@@ -307,6 +307,17 @@ def values(self):
...
 
 
+def _safe_open_write(module, path, follow):
+    flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+    if not follow and hasattr(os, 'O_NOFOLLOW'):
+        flags |= os.O_NOFOLLOW
+    try:
+        fd = os.open(path, flags, int('0600', 8))
+    except OSError as e:
+        module.fail_json(msg="Refusing to operate on symlink: %s : %s" % (path, to_native(e)))
```
bcoca: there is more than symlink errors possible at this point, I'm not sure this function is needed at all, but if we insist on adding it, the error should reflect all possible issues.

plugins/modules/authorized_key.py
```diff
@@ -359,25 +370,25 @@ def keyfile(module, user, write=False, path=None, manage_dir=True, follow=False)
...
-        f = None
-        try:
-            f = open(keysfile, "w")  # touches file so we can set ownership and perms
-        finally:
-            f.close()
+        fd = _safe_open_write(module, keysfile, follow)
+        os.close(fd)
         if module.selinux_enabled():
             module.set_default_selinux_context(keysfile, False)
+    elif not follow and os.path.islink(keysfile):
```
bcoca: this is a change in behavior, not sure how it applies to the fix

bcoca (dismissed):

plugins/modules/authorized_key.py
saito-hideki:
```python
    except OSError as e:
      module.fail_json(msg="File open failed %s : %s" % (path, to_native(e)))
```
I will change it to a more general error message like above.
Since the details should be displayed by `to_native(e)`, I think this fixes the confusing message.

plugins/modules/authorized_key.py
saito-hideki: I will remove that `elif` block.
I also noticed that on Linux, `os.chmod()` does not support `follow_symlinks`(macOS is OK).

```
~~~
fatal: [server00.fusion.fgrep.org]: FAILED! => {"changed": false, "msg": "Task failed: Module failed: chmod: follow_symlinks unavailable on this platform"}
~~~

> RHEL9
$ uname -sr
Linux 5.14.0-687.15.1.el9_8.aarch64
$ python -c "import os;[print(x) for x in os.supports_follow_symlinks]"
<built-in function chown>
<built-in function access>
<built-in function link>
<built-in function utime>
<built-in function stat>

> macOS 26.5.1
% uname -sr
Darwin 25.5.0
% python -c "import os;[print(x) for x in os.supports_follow_symlinks]"
<built-in function chown>
<built-in function utime>
<built-in function chmod>   # HERE
<built-in function stat>
<built-in function chflags>
<built-in function link>
<built-in function access>
```

Since CVE-2026-11837 is regarding `os.chown()`. Therefore, I will revert `os.chmod()` to its original implementation so that it follows the traditional behavior.
If you have any concerns regarding this matter, please let me know.

saito-hideki:
Hi @bcoca @Akasurde @gundalow @Andersson007 
Thank you for your help with this PR!
I've addressed the change request from @boca. Could you please review it again?
I think I've fixed the implementation that would change the existing behavior.

The revision policy is as follows
1. Keep existing behavior, except for CVE-2026-11837 fixes (as advised by @bcoca )
2. To improve future reusability, extract the file-writing logic that implements the fix in point 1. into a separate function(as advised by @Akasurde)
These are the two points.

Also, I noticed that os.chown in keyfile() runs before writefile() and atomic_move(), so its effect is lost when atomic_move replaces the symlink with a real file.
So I have added the following block to set the correct ownership and permissions after the file replacement, to maintain the same behavior as in 2.2.0. I don't think the following block affects CVE-2026-11837.

- https://github.com/ansible-collections/ansible.posix/pull/760/changes#diff-e11f7706a73049ab9194f2dc8f7db4b1b1643e6c129ad174244a02b8a35e3842R707-R711

Andersson007 (approved):

saito-hideki:
@Andersson007 Thank you for approving it!
@Akasurde @bcoca @gundalow If everyone is OK with it, I plan to merge this and release it as version 2.2.1 early next week. Please let me know your thoughts.

saito-hideki:
Since there don't seem to be any issues, I'll merge this and backport it to stable-1 and stable-2.
After that, I plan to begin work on the 2.2.1 and 1.6.3 releases.

PR #773, opened 2026-07-22
fix(callback): align profile_tasks recap output width

aioue:
## Summary

- `profile_tasks`: replace the hardcoded 79-column `filled()` width with `display.columns`, matching `display.banner()` fill rules (`columns + 1` visible width). Removes the trailing space `filled()` appended after the stars, which left the last `*` one column left of the TASK banner.
- `profile_tasks`: add `truncate_task_names` (default `false`) to ellipsize long task names in the TASKS RECAP summary. Enable with `PROFILE_TASKS_TRUNCATE_TASK_NAMES` or `callback_profile_tasks.truncate_task_names`.

The `ansible.posix.debug` PLAY RECAP alignment from an earlier revision of this PR is now in #778.

## Output

Captured from `playbooks/test/callback-plugins.yml` (80-column terminal, `columns=79`).

### Per-task timestamp (during play)

The visible lines below look almost identical in copy/paste. The misalignment is a single invisible trailing space on the timestamp line in 2.2.2. See line endings below.

Before (`ansible.posix` 2.2.2):

```
TASK [Short task] **************************************************************
Wednesday 22 July 2026  15:25:24 +0000 (0:00:01.242)       0:00:01.266 ********
```

After:

```
TASK [Short task] **************************************************************
Wednesday 22 July 2026  15:25:27 +0000 (0:00:01.228)       0:00:01.244 *********
```

Line endings (`cat -A` on the timestamp line only):

```
# before: eight stars, then a space, then end of line
... 0:00:01.266 ******** $

# after: nine stars, then end of line
... 0:00:01.244 *********$
```

### TASKS RECAP

Before: header/separator are 80 columns with a trailing space; timing rows are 79 columns. Long task names overflow.

```
TASKS RECAP ********************************************************************
Wednesday 22 July 2026  15:25:24 +0000 (0:00:00.382)       0:00:01.670 ********
=============================================================================== 
Duplicate --------------------------------------------------------------- 1.24s
This is a deliberately long task name used to test TASKS RECAP truncation behaviour --- 0.38s
```

After: all rows are 80 columns. Long names still print in full by default. With `truncate_task_names=true` they are ellipsized.

```
TASKS RECAP ********************************************************************
Wednesday 22 July 2026  15:25:27 +0000 (0:00:00.407)       0:00:01.681 *********
================================================================================
Duplicate ---------------------------------------------------------------- 1.23s
This is a deliberately long task name used to test TASKS RECAP truncation behaviour --- 0.41s
```

## Test plan

- [x] Run a playbook with `callbacks_enabled=ansible.posix.profile_tasks`
- [x] Verify TASK banner and profile_tasks timestamp lines end on the same star column (`cat -A` on the timestamp line)
- [x] Verify TASKS RECAP header, timestamp, separator, and timing rows are all the same width
- [x] Verify long task names are not truncated unless `truncate_task_names` is enabled
- [x] Verify duplicate-named tasks appear separately in TASKS RECAP (UUID tracking unchanged)

### DIFF
--- a/plugins/callback/profile_tasks.py
+++ b/plugins/callback/profile_tasks.py
@@ -63,6 +63,17 @@ DOCUMENTATION = '''
           - section: callback_profile_tasks
             key: datetime_format
         version_added: 3.0.0
+      truncate_task_names:
+        description:
+          - Truncate long task names in the TASKS RECAP summary with an ellipsis so rows stay within the terminal width.
+        type: bool
+        default: false
+        env:
+          - name: PROFILE_TASKS_TRUNCATE_TASK_NAMES
+        ini:
+          - section: callback_profile_tasks
+            key: truncate_task_names
+        version_added: 2.3.0
 '''
 
 EXAMPLES = '''
@@ -88,6 +99,7 @@ from datetime import datetime
 
 from functools import reduce
 from ansible.plugins.callback import CallbackBase
+from ansible.utils.display import get_text_width
 
 
 # define start time
@@ -102,16 +114,64 @@ def secondsToStr(t):
     return "%d:%02d:%02d.%03d" % tuple(reduce(rediv, [[t * 1000, ], 1000, 60, 60]))
 
 
-def filled(msg, fchar="*"):
-    if len(msg) == 0:
-        width = 79
+def filled(msg, columns, fchar="*"):
+    """Fill to the same width as display.banner() (columns + 1)."""
+    if not msg:
+        return fchar * (columns + 1)
+    try:
+        msg_width = get_text_width(msg)
+    except EnvironmentError:
+        msg_width = len(msg)
+    fill_len = columns - msg_width
+    if fill_len < 3:
+        fill_len = 3
+    return "%s %s" % (msg, fchar * fill_len)
+
+
+def _text_width(text):
+    try:
+        return get_text_width(text)
+    except EnvironmentError:
+        return len(text)
+
+
+def truncate_name(name, width):
+    if width <= 0:
+        return name
+    if _text_width(name) <= width:
+        return name
+    if width <= 3:
+        return name[:width]
+    ellipsis = '...'
+    for end in range(len(name), 0, -1):
+        candidate = name[:end] + ellipsis
+        if _text_width(candidate) <= width:
+            return candidate
+    return ellipsis
+
+
+def pad_to_display_width(text, width, fillchar='-'):
+    current = _text_width(text)
+    if current >= width:
+        return text
+    return text + fillchar * (width - current)
+
+
+def format_timing_row(task_name, elapsed, columns, truncate=False):
+    line_width = columns + 1
+    time_field_width = 9
+    name_field_width = line_width - time_field_width
+    if truncate:
+        task_name = truncate_name(task_name, name_field_width - 1)
+    prefix = task_name + u' '
+    gap = name_field_width - _text_width(prefix)
+    if gap == 1:
+        # A lone dash before the timing field looks like misaligned padding.
+        left = prefix + u' '
     else:
-        msg = "%s " % msg
-        width = 79 - len(msg)
-    if width < 3:
-        width = 3
-    filler = fchar * width
-    return "%s%s " % (msg, filler)
+        left = pad_to_display_width(prefix, name_field_width)
+    timing = u' {0:.02f}s'.format(elapsed)
+    return left + timing.rjust(time_field_width, u'-')
 
 
 def timestamp(self):
@@ -127,7 +187,10 @@ def tasktime(self):
     time_elapsed = secondsToStr((cdtn - dtn).total_seconds())
     time_total_elapsed = secondsToStr((cdtn - dt0).total_seconds())
     dtn = cdtn
-    return filled('%s (%s)%s%s' % (datetime_current, time_elapsed, ' ' * 7, time_total_elapsed))
+    return filled(
+        '%s (%s)%s%s' % (datetime_current, time_elapsed, ' ' * 7, time_total_elapsed),
+        self._display.columns,
+    )
 
 
 class CallbackModule(CallbackBase):
@@ -148,6 +211,7 @@ class CallbackModule(CallbackBase):
         self.summary_only = None
         self.task_output_limit = None
         self.datetime_format = None
+        self.truncate_task_names = None
 
         super(CallbackModule, self).__init__()
 
@@ -177,6 +241,8 @@ class CallbackModule(CallbackBase):
         if self.datetime_format is not None:
             if self.datetime_format == 'iso8601':
                 self.datetime_format = '%Y-%m-%dT%H:%M:%S.%f'
+
+        self.truncate_task_names = self.get_option('truncate_task_names')
 
     def _display_tasktime(self):
         if not self.summary_only:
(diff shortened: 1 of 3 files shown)


aioue:
@saito-hideki code has drifted, have rebased onto main. Let me know if there is anything else you need to get this merged, please? Thank you 🙏

plugins/callback/debug.py
saito-hideki, line 1: Since `ansible.posix.debug` and `ansible.posix.profile_tasks` are separate plugins, I propose splitting them into separate pull requests.
I think this pull request #773 should focus on `profile_tasks`.

plugins/callback/profile_tasks.py
```diff
@@ -63,6 +63,17 @@
           - section: callback_profile_tasks
             key: datetime_format
         version_added: 3.0.0
+      truncate_task_names:
+        description:
+          - Truncate long task names in the TASKS RECAP summary with an ellipsis so rows stay within the terminal width.
+        type: bool
+        default: true
```
saito-hideki: For backward compatibility, I think it's better to set the default value to `false`.
This prevents task names from being truncated unintentionally.

plugins/callback/profile_tasks.py
```diff
@@ -63,6 +63,17 @@
...
+        env:
+          - name: PROFILE_TASKS_TRUNCATE_TASK_NAMES
+        ini:
+          - section: callback_profile_tasks
+            key: truncate_task_names
+        version_added: 3.1.0
```
saito-hideki: Since version 3.0.0 hasn't been released yet, this is fine for version 3.0.0 :)

saito-hideki (changes requested):
@aioue thanks for the PR!
I left a few comments on that review. My main suggestion is that the fix for `ansible.posix.debug` should be split off into a separate pull request :)
Please let us know your thoughts.

plugins/callback/debug.py
aioue: Split into #778 so this PR stays on `profile_tasks` only.

plugins/callback/profile_tasks.py
aioue: Done: default is now `false`.

plugins/callback/profile_tasks.py
aioue: Updated to `3.0.0`.

aioue:
@saito-hideki thanks for the review. I split the `ansible.posix.debug` PLAY RECAP change into #778 so this PR is `profile_tasks` only.

`truncate_task_names` now defaults to `false`, and `version_added` is `3.0.0`.

saito-hideki:
Hi @aioue 

I ran a few tests.
This fix does not seem to work correctly when handling multibyte characters, such as in CJK environment.
I'll try to come up with a solution, too.

- ansible.cfg:

```ini
% cat ansible.cfg
[defaults]
callbacks_enabled = ansible.posix.profile_tasks
collections_path =./collections
host_key_checking = False

[callback_profile_tasks]
truncate_task_names = true
```
- Playbook:

```
---
- hosts: localhost
  gather_facts: false
  tasks:
    - name: "01234567890123456789012345678901234567890123456789012345678901234567890123456789"
      ansible.builtin.debug:
        msg: "ok"
    - name: "零一二三四五六七八九零一二三四五六七八九零一二三四五六七八九零一二三四五六七八九"
      ansible.builtin.debug:
        msg: "ok"
...
```

- Playbook Execution Results:

```shell
% ansible-playbook tests/playbook/pr_773.yml
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] ***************************************************************

TASK [01234567890123456789012345678901234567890123456789012345678901234567890123456789] ***
Tuesday 18 August 2026  17:30:32 +0900 (0:00:00.004)       0:00:00.004 *********
ok: [localhost] => {
    "msg": "ok"
}

TASK [零一二三四五六七八九零一二三四五六七八九零一二三四五六七八九零一二三四五六七八九] ***
Tuesday 18 August 2026  17:30:32 +0900 (0:00:00.010)       0:00:00.015 *********
ok: [localhost] => {
    "msg": "ok"
}

PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

TASKS RECAP ********************************************************************
Tuesday 18 August 2026  17:30:32 +0900 (0:00:00.009)       0:00:00.024 *********
================================================================================
0123456789012345678901234567890123456789012345678901234567890123456... --- 0.01s
零一二三四五六七八九零一二三四五六七八九零一二三四五六七八九零一二三四五六七八九 --------------------------------- 0.01s
```

aioue:
@saito-hideki Thanks for the CJK reproducer. It showed two bugs in the truncation path.

`truncate_name()` compared `len()` to the column budget. Your 40-character task name is under 71 characters but uses 80 terminal columns, so it was never ellipsized.

Recap rows also padded with Python `str.format` character width, so wide characters broke alignment even when the name was short enough.

Both paths now use `get_text_width()` (the same helper `display.banner()` uses). I kept dash fill on the timing field to match the original `{1:->9}` layout. Unit tests cover long ASCII and CJK names, including your case.

Happy to run anything else you want checked.

aioue:
Added unit tests for the CJK reproducer from https://github.com/ansible-collections/ansible.posix/pull/773#issuecomment-5325694910:

- 40-character CJK name that fits under the character budget but not the display-width budget
- Truncation with ellipsis for that name
- ASCII and CJK recap rows both at 80 terminal columns
- Dash padding on the timing field
- `pad_to_display_width()` with CJK input

Pushed in 9e06481.

aioue:
**With truncate (80 char width terminal)**

image

**Without truncate**

image

---
If you could kindly review [the companion PR](https://github.com/ansible-collections/ansible.posix/pull/778), please.

plugins/callback/profile_tasks.py
```diff
@@ -63,6 +63,17 @@
...
+        description:
+          - Truncate long task names in the TASKS RECAP summary with an ellipsis so rows stay within the terminal width.
+        type: bool
+        default: false
+        env:
+          - name: PROFILE_TASKS_TRUNCATE_TASK_NAMES
+        ini:
+          - section: callback_profile_tasks
+            key: truncate_task_names
+        version_added: 3.0.0
```
saito-hideki: Since we set the default value of `truncate_task_names` to `false`, there are no longer any major changes to the default behavior.
So I have determined that this can be treated as version 2.3.0 for now.
It was previously updated to 3.0.0, but could you please update it to 2.3.0? Since this is a valid bug fix, I plan to incorporate it into `stable-2` so that it can be included in the next 2.3.0 release.

Sorry for making you do it twice.

saito-hideki (changes requested):
@aioue 
Sorry for the delay!
Just one thing, could you change `version_added: 3.0.0` to `2.3.0`? 
I'd like to incorporate your fix into `stable-2` as well.

plugins/callback/profile_tasks.py
aioue: Done. Thank you very much for your time and patience with this PR, I appreciate your help and expertise.

saito-hideki (approved):
LGTM.
Thanks! :)

aioue:
@saito-hideki Thanks again for the review.

`truncate_task_names` is now marked `version_added: 2.3.0` as requested.

CI is green on everything relevant to this PR (sanity, units, lint across Ansible 2.16 through devel, Docker, RHEL 9.x, FreeBSD, SonarCloud, and Zuul). The only failures are the Remote RHEL 10 jobs, which die during Python bootstrap before any collection tests run:

```
nothing provides python3 = 3.12.14-1.el10_2 needed by python3-devel-3.12.14-1.el10_2.x86_64
```

That looks like a CI image/repo mismatch rather than anything in this change. Happy to re-run or adjust if you see anything else needed.

saito-hideki:
@aioue Indeed.
I'll make time tomorrow to double-check the CI image issue :)

saito-hideki:
recheck

saito-hideki:
@aioue The lint checker in the GitHub workflow was failing, but since this is a CI issue, I will handle it separately. Of course, there’s no problem with the PR itself, so I proceeded with the merge process.
Thank you for your contribution!

PR #774, opened 2026-08-05
mount: Create a state like absent but keeps the mountpoint

### LINKED ISSUE #331: Add option to not remove mountpoint
##### SUMMARY
It would be useful to add the option to preserve the mountpoint directory when the state is set to `absent`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
ansible.posix.mount (plugins/modules/mount.py)

##### ADDITIONAL INFORMATION
I use network mounts over certain directories that varies from host to host. Nearly every time I specify the state as `absent`, it's because I want to ensure that a host is using the native filesystem rather than a network share. I want the underlying directory to still exist and rather not have to use a `file` task.

EDIT: I'm happy to do a PR don't have a good name for the option. It could also double as a flag to not have this task create directories needed for the mount.

sprottm:
##### SUMMARY

- Currently, `state: absent` will remove the mount, the entry in fstab, and the mountpoint
  - Users might not want to run rmdir on the mountpoint. Maybe they mounted over something else. 
- We want to keep backwards compatibility.
  - If we added a new option for removing a mountpoint, it would end up changing how an existing state works.
  - By adding a new state, users can just opt into the new state.
    - I decided to call the state `absent_mount`.
    - `absent` here means that the mountpoint, fstab entry, and mount are all removed
    - `absent_mount` means that only the mount is removed (to include the fstab entry), and the mountpoint stays  
- Fixes #331 
- Addresses same concerns as #569 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

mount

##### ADDITIONAL INFORMATION

```paste below
# See integration tests
```

### DIFF
--- a/plugins/modules/mount.py
+++ b/plugins/modules/mount.py
@@ -88,6 +88,13 @@ options:
         fail if multiple devices are mounted on the same mount point. Using
         V(absent) with a mount point that is not registered in the I(fstab) has
         no effect, use V(unmounted) instead.
+      - V(absent_mount) specifies that the mount point entry O(path) will be removed
+        from I(fstab) and will also unmount the mounted device. The associated
+        mountpoint will not be removed. A mounted device will be unmounted regardless 
+        of O(src) or its real source. V(absent_mount) does not unmount recursively, 
+        and the module will fail if multiple devices are mounted on the same mount
+        point. Using V(absent_mount) with a mount point that is not registered in the
+        I(fstab) has no effect, use V(unmounted) instead.
       - V(remounted) specifies that the device will be remounted for when you
         want to force a refresh on the mount itself (added in 2.9). This will
         always return RV(ignore:changed=true). If O(opts) is set, the options will be
@@ -102,7 +109,7 @@ options:
         mountpoint.
     type: str
     required: true
-    choices: [ absent, absent_from_fstab, mounted, present, unmounted, remounted, ephemeral ]
+    choices: [ absent, absent_from_fstab, absent_mount, mounted, present, unmounted, remounted, ephemeral ]
   fstab:
     description:
       - File to use instead of C(/etc/fstab).
@@ -169,6 +176,11 @@ EXAMPLES = r'''
   ansible.posix.mount:
     path: /tmp/mnt-pnt
     state: unmounted
+
+- name: Unmount a mounted volume and remove its fstab entry
+  ansible.posix.mount:
+    path: /tmp/mnt-pnt
+    state: absent_mount
 
 - name: Remount a mounted volume
   ansible.posix.mount:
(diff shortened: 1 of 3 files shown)


plugins/modules/mount.py
```diff
@@ -88,6 +88,13 @@
         fail if multiple devices are mounted on the same mount point. Using
         V(absent) with a mount point that is not registered in the I(fstab) has
         no effect, use V(unmounted) instead.
+      - V(absent_mount) specifies that the mount point entry O(path) will be removed
```
saito-hideki: I think that it's hard to imagine the behavior of the option named `absent_mount` just from the option name.
How about renaming it to something like `absent_keep_mountpoint` or `absent_exclude_mountpoint`?

saito-hideki (changes requested):
@sprottm thanks for the PR!
Based on the name `absent_mount` I think it’s hard to imagine how this option works.
How about renaming it to something like `absent_keep_mountpoint` or `absent_exclude_mountpoint` ?
Since there is already an option name consisting of three words`absent_from_fstab`. So I don't think it necessarily has to be two words. Please let me know what you think.

plugins/modules/mount.py
sprottm: I renamed it to absent_keep_mountpoint since I prefer the sound of that.

sprottm:
Alright, well the failed pipelines have nothing to do with my code.

I can see about MAYBE helping to resolve these issues on a separate PR when I get home from work; assuming someone else hasn't already started on it by the time that I get back.

```
01:52 Error: 
01:52  Problem: cannot install the best candidate for the job
01:52   - nothing provides python3 = 3.12.14-1.el10_2 needed by python3-devel-3.12.14-1.el10_2.x86_64 from rhel-10-appstream-rhui-rpms
01:52   - nothing provides python3-libs(x86-64) = 3.12.14-1.el10_2 needed by python3-devel-3.12.14-1.el10_2.x86_64 from rhel-10-appstream-rhui-rpms
01:52 Failed to install packages. Giving up.
```

plugins/modules/mount.py
```diff
@@ -881,7 +906,7 @@ def main():
 
     if state == 'absent_from_fstab':
         name, changed = unset_mount(module, args)
-    elif state == 'absent':
+    elif state == 'absent' or state == 'absent_keep_mountpoint':
         name, changed = unset_mount(module, args)
```
saito-hideki: It seems there are conflicts due to #742. If possible, could you please rebase using the latest main branch?
```
    if state == 'absent_from_fstab':
<<<<<<< HEAD
        name, changed, _dummy = unset_mount(module, args, backup=False)
    elif state == 'absent':
        name, changed, backup_lines = unset_mount(module, args, backup=True)
=======
        name, changed = unset_mount(module, args)
    elif state == 'absent' or state == 'absent_mount':
        name, changed = unset_mount(module, args)
>>>>>>> 8b581c1 (Create a new state, with an example and integration test)
```

saito-hideki (changes requested):
Thanks for the fix. It looks like a rebase is needed due to the merged of #742.
Please rebase using the latest main branch and resolve any conflicts.

saito-hideki (approved):
LGTM!
Thanks for your contribution.

saito-hideki:
@sprottm, your PR has been merged.
Thanks!

PR #779, opened 2026-08-18
ACA-4558: synchronize: use ansible_host from play vars when absent from hostvars

### LINKED ISSUE #469: sycnronise uses `inventory_name` and not `ansible_host` for rsync connection
### Summary

To keep inventories simpler I've specified hosts by hostname (i.e. `host1`, `host2`, etc.) and set the `ansible_host` variable to `"{{ inventory_hostname }}.domain.org"`, or in a task that checks if a hosts DNS entry is valid where I set `ansible_host` to an IP address.

However when I use `ansible.builtin.synchronise` it uses a hosts `inventory_name` and not `ansible_host` to do the rsync operation, which leads to "Host not found errors" if the target host has a doman that is not the same as the localhost, and not in the host's search domains.

My expectation would be that the `ansible.builtin.synchronize` module would use `ansible_host` to do its rsync operations, just as every other ansible module does.

### Issue Type

Bug Report

### Component Name

ansible.builtin.synchronize

### Ansible Version

```console
> ansible --version
ansible [core 2.13.8]
  config file = /home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg
  configured module search path = ['/home/hicksaw/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/hicksaw/ansible/gitlab-15-ansible/venvs/ansible/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/hicksaw/ansible/gitlab-15-ansible/ignore/collections
  executable location = /home/hicksaw/ansible/gitlab-15-ansible/venvs/ansible/bin/ansible
  python version = 3.9.13 (main, Nov  9 2022, 13:16:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)]
  jinja version = 3.1.2
  libyaml = True
```

### Configuration

```console
# if using a version older than ansible-core 2.12 you should omit the '-t all'
> ansible-config dump --only-changed
CACHE_PLUGIN(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = ignore/facts_cache
CALLBACKS_ENABLED(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = ['profile_tasks', 'log_plays']
COLLECTIONS_PATHS(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = ['/home/hicksaw/ansible/gitlab-15-ansible/ignore/collections']
DEFAULT_GATHERING(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = smart
DEFAULT_LOAD_CALLBACK_PLUGINS(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = True
DEFAULT_ROLES_PATH(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = ['/home/hicksaw/ansible/gitlab-15-ansible/ignore/roles']
DEFAULT_STDOUT_CALLBACK(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = yaml
DEFAULT_TIMEOUT(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = 60
DEFAULT_VAULT_IDENTITY_LIST(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = ['.vault_password']
DISPLAY_SKIPPED_HOSTS(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = False
HOST_KEY_CHECKING(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = False
INVENTORY_CACHE_ENABLED(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = True
INVENTORY_CACHE_PLUGIN_CONNECTION(/home/hicksaw/ansible/gitlab-15-ansible/ansible.cfg) = ignore/inventory_cache
```

### OS / Environment

Rocky 8.4

### Steps to Reproduce

```yaml (paste below)
- name: Sync files
  hosts: all
  vars:
    ansible_host: "{{ inventory_name }}.domain.org"

  task:
  - name: Syncronise files
      ansible.builtin.synchronize:
        src: "/tmp/sync_dir/"
        dest: /tmp/sync_dir/
        delete: true
      become: true
```

### Expected Results

I expected the `ansible.builtin.syncronize` to use `ansible_host` to do an rsync operation.

### Actual Results

```console
TASK [Sync files] ****************************************************************************************************************************************************
fatal: [hostq]: FAILED! => changed=false 
  cmd: sshpass -d4 /usr/bin/rsync --delay-updates -F --compress --delete-after --archive --no-owner --no-group --rsh='/usr/bin/ssh -S none -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null' --out-format='<<CHANGED>>%i %n%L' user@host1:/tmp/sync_dir/ /tmp/sync_dir/
  msg: |-
    sss_ssh_knownhostsproxy: Could not resolve hostname host1
    kex_exchange_identification: Connection closed by remote host
    rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
    rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.3]
  rc: 255
```

Copypaste from: https://github.com/ansible/ansible/issues/80934

sussen-redhat:
Fixes #469.

## Problem

`synchronize` reads `ansible_host` only from `task_vars['hostvars'][inventory_hostname]`. When `ansible_host` is set via play-level `vars:` instead of inventory/group_vars/host_vars, it is absent from that cached snapshot and the plugin silently falls back to bare `inventory_hostname`, producing 'Host not found' errors.

Confirmed reproduction: with `ansible_host: "{{ inventory_hostname }}.domain.org"` in play `vars:`, the rsync destination was built as `user@host1:/path` instead of `user@host1.domain.org:/path`.

## Fix

- Fall back to `task_vars['vars'].get('ansible_host')` (the play variable namespace) when `hostvars[inventory_hostname]` lacks `ansible_host`. This correctly handles play-level `vars:` without being affected by `delegate_to` context (which clobbers top-level `task_vars['ansible_host']` with the delegate host's value).
- Explicitly template `dest_host` since play vars can be Jinja expressions.
- Extend the push-mode password lookup with the same `task_vars` fallback.

## Test fixes

Also updates `TaskMock` and `ConnectionMock` with attributes added in ansible-core 2.15 (`always_pipeline_modules`, `become`, `has_pipelining`, `socket_path`, `diff`, `delegate_to`, `no_log`) and fixes class-level state leaking between `SynchronizeTester` test runs by resetting `task` and `connection` at the start of each `runtest()` call.

## Testing

- Added `fixtures/synchronize/ansible_host_from_task_vars/` and `test_ansible_host_from_task_vars`.
- Confirmed the test **fails** on unpatched code (`dest: root@el6host:/tmp/deleteme`) and **passes** with the fix (`dest: root@el6host.example.org:/tmp/deleteme`).
- Full suite: **13/13 passed**.


tests/unit/plugins/action/fixtures/synchronize/ansible_host_from_task_vars/taskvars_in.json
```diff
@@ -0,0 +1,150 @@
...
+  "inventory_hostname_short": "el6host",
+  "omit": "__omit_place_holder__03600813b83569c710bf5cb2a040d6e01da927c6",
+  "inventory_dir": "/home/jtanner/workspace/issues/AP-15905",
+  "playbook_dir": "/home/jtanner/workspace/issues/AP-15905",
+  "role_names": [],
+  "play_hosts": [
+    "el6host"
+  ],
+  "ansible_sftp_extra_args": ""
+}
```
saito-hideki: My understanding is that this file was created based on https://github.com/ansible-collections/ansible.posix/blob/main/tests/unit/plugins/action/fixtures/synchronize/basic/taskvars_in.json.

However, if we’re focusing solely on this issue, wouldn’t something as simple as the following be sufficient?

```JSON
{
  "inventory_hostname": "el6host",
  "ansible_user": "root",
  "hostvars": {
    "el6host": {}
  },
  "vars": {
    "ansible_host": "el6host.example.org"
  }
}
```

tests/unit/plugins/action/fixtures/synchronize/ansible_host_from_task_vars/meta.yaml
```diff
@@ -0,0 +1,16 @@
+---
+fixtures:
```
saito-hideki, line 6: Since you can include comments in `meta.yaml`, please include a description of the test's purpose as a comment if possible.

plugins/action/synchronize.py
```diff
@@ -240,7 +240,11 @@ def run(self, tmp=None, task_vars=None):
         src_host = '127.0.0.1'
         inventory_hostname = task_vars.get('inventory_hostname')
         dest_host_inventory_vars = task_vars['hostvars'].get(inventory_hostname)
-        dest_host = dest_host_inventory_vars.get('ansible_host', inventory_hostname)
+        dest_host = (dest_host_inventory_vars.get('ansible_host')
```
saito-hideki: AFIK, `task_vars.get('vars', {}).get('ansible_host')` ... this “vars” feature is scheduled to be deprecated in Ansible Core 2.24.
- https://github.com/ansible/ansible/blob/devel/lib/ansible/vars/manager.py#L57-L62

Also, since `task_vars` should take precedence, I think the order will be as follows:

```python
  dest_host = (task_vars.get('ansible_host')
               or dest_host_inventory_vars.get('ansible_host')
               or inventory_hostname)
```

plugins/action/synchronize.py
```diff
@@ -240,7 +240,11 @@ def run(self, tmp=None, task_vars=None):
...
+                     or task_vars.get('vars', {}).get('ansible_host')
+                     or inventory_hostname)
+        if self._templar is not None:
+            dest_host = self._templar.template(dest_host)
 
         dest_host_ids = [hostid for hostid in (dest_host_inventory_vars.get('inventory_hostname'),
```
saito-hideki, line 251: I think we should add potential hosts for `delegate_to` in this case.

```python
        dest_host_ids = [hostid for hostid in (dest_host_inventory_vars.get('inventory_hostname'),
                                               dest_host_inventory_vars.get('ansible_host'),
                                               dest_host)
                         if hostid is not None]
```

plugins/action/synchronize.py
```diff
@@ -371,7 +375,10 @@ def run(self, tmp=None, task_vars=None):
                 src = self._process_origin(src_host, src, user)
                 dest = self._process_remote(_tmp_args, dest_host, dest, user, inv_port in localhost_ports)
 
-            password = dest_host_inventory_vars.get('ansible_ssh_pass', None) or dest_host_inventory_vars.get('ansible_password', None)
+            password = (dest_host_inventory_vars.get('ansible_ssh_pass', None)
```
saito-hideki: We need to prioritize `task_vars` in the same order as `dest_host`.

```python
            password = (task_vars.get('ansible_ssh_pass', None)
                        or task_vars.get('ansible_password', None)
                        or dest_host_inventory_vars.get('ansible_ssh_pass', None)
                        or dest_host_inventory_vars.get('ansible_password', None))
```

saito-hideki (changes requested):
@sussen-redhat Thank you for the PR.
Overall, I agree this proposed approach. I have included comments on the specific points among the proposed changes that caught my attention, so I would appreciate it if you could review them and take appropriate action.

tests/unit/plugins/action/fixtures/synchronize/ansible_host_from_task_vars/taskvars_in.json
sussen-redhat: Done — simplified to a minimal fixture as suggested.

tests/unit/plugins/action/fixtures/synchronize/ansible_host_from_task_vars/meta.yaml
sussen-redhat: Done — added a comment describing the test purpose.

plugins/action/synchronize.py
sussen-redhat: Done — changed to task_vars.get('ansible_host') first, guarded with delegate_to is None to avoid picking up the delegate host's address in delegation scenarios where top-level task_vars['ansible_host'] reflects the delegate rather than the inventory target.

plugins/action/synchronize.py
sussen-redhat: Done — added dest_host to dest_host_ids.

plugins/action/synchronize.py
sussen-redhat: Done — task_vars now takes precedence over dest_host_inventory_vars.

sussen-redhat:
All four points from the review have been addressed in the second commit (a9f86cf). All 13 unit tests pass. Thank you for the thorough review.

plugins/action/synchronize.py
saito-hideki: Please set `task_vars` as the highest priority and add a comment regarding the `delegate_to` guard. It would look something like this, for example.
The comment in the sample is based on my understanding of this implementation. Please feel free to correct it if you have a better explanation.

```python
          # If the user did not set ansible_host, ansible fills it in with the address
          # of the delegate host when delegating, so only use ansible_host when not delegating.
          dest_host = ((task_vars.get('ansible_host') if delegate_to is None else None)
                       or dest_host_inventory_vars.get('ansible_host')
                       or inventory_hostname)
```

sussen-redhat: Thank you for the suggestion. I've updated the second commit (18b6293) with the following changes:

- Added the comment explaining the delegate_to guard above the dest_host assignment, using the wording you suggested.
- Moved task_vars.get('ansible_host') to the highest priority as requested, keeping the delegate_to is None guard.
- Applied the same pattern consistently to the password lookup (ansible_ssh_pass/ansible_password) since it has the same delegation issue — when delegating, Ansible fills in task_vars with the delegate host's values, so we should only use task_vars when not delegating.

- The final dest_host logic now reads:

```
python
# If the user did not set ansible_host, ansible fills it in with the address
# of the delegate host when delegating, so only use ansible_host when not delegating.
dest_host = ((task_vars.get('ansible_host') if delegate_to is None else None)
             or dest_host_inventory_vars.get('ansible_host')
             or inventory_hostname)
```

All 13 unit tests pass.

plugins/action/synchronize.py
```diff
@@ -371,7 +378,13 @@ def run(self, tmp=None, task_vars=None):
                 src = self._process_origin(src_host, src, user)
                 dest = self._process_remote(_tmp_args, dest_host, dest, user, inv_port in localhost_ports)
 
-            password = dest_host_inventory_vars.get('ansible_ssh_pass', None) or dest_host_inventory_vars.get('ansible_password', None)
+            # If the user did not set ansible_ssh_pass/ansible_password, ansible fills it in
+            # with the delegate host's value when delegating, so only use task_vars when not
+            # delegating.
+            password = ((task_vars.get('ansible_ssh_pass') if delegate_to is None else None)
```
saito-hideki: With delegate_to set and ansible_ssh_pass given in task_vars, this guard drops the target password and falls back to the stale hostvars value, so rsync fails to authenticate.
In this case, password is never injected the way ansible_host is, so please drop the guard here.

I think it would be better to remove delegate_to guards.

```python
              password = (task_vars.get('ansible_ssh_pass')
                          or task_vars.get('ansible_password')
                          or dest_host_inventory_vars.get('ansible_ssh_pass', None)
                          or dest_host_inventory_vars.get('ansible_password', None))
```

Above code is working properly on my end.

saito-hideki (changes requested):
I have left a comment. Please double-check password logic.

plugins/action/synchronize.py
sussen-redhat: Done — removed the delegate_to guard from the password lookup as suggested. You're right that unlike ansible_host, Ansible does not inject the delegate host's credentials into task_vars for ansible_ssh_pass/ansible_password — those always refer to the target host's credentials — so the guard was incorrect and would drop the password when delegating. The final password logic now matches your suggestion exactly:

```
python
password = (task_vars.get('ansible_ssh_pass')
            or task_vars.get('ansible_password')
            or dest_host_inventory_vars.get('ansible_ssh_pass', None)
            or dest_host_inventory_vars.get('ansible_password', None))
```

All 13 unit tests pass.

saito-hideki (approved):
@sussen-redhat LGTM.
Thanks!

PR #790, opened 2026-09-07
Draft for Issue #256, feature for adding comment at the top of the file

### LINKED ISSUE #256: ansible.posix.authorized_key: allow setting a comment in authorized keys file (with #, not at end of each key)
##### SUMMARY
I want to insert a comment in authorized_keys like
```
# NOTE: this authorized_keys file is managed by Ansible - DO NOT MODIFY IT MANUALLY, CONTACT ADMIN@EMAIL INSTEAD
```
to avoid users trying to manually changing their keys and getting confused when they are overwritten by our centrally managed key management with Ansible.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
ansible.posix.authorized_key

##### ADDITIONAL INFORMATION
 In particular, we want to avoid spurious key changes (users manually editing by accident) while remaining sensitive to key changes happening for other reasons for security purposes (e.g. if there is a security breach and an attacker modifies the keys we want to see that ansible has detected changes in the keys), so this rules out non-idempotent approaches to add a comment (lineinfile) and changed_when: false.

Also, if users modify their authorized_keys file, say to put in a new key they made and remove an old one, then if we run Ansible it will put the old one back and they might lose login access.

Currently comments can only be added as a field at the end of each key, which has a completely different purpose (identifying each key).

It seems [in the distant past](https://github.com/ansible/ansible/issues/521) Ansible automatically inserted a comment in the file, but not anymore.  Would be good if it is optional and configurable.

I even tried inserting a dummy key "# COMMENT" , however due to https://github.com/ansible/ansible/pull/5655/  lines beginning with `#` are ignored so it does not get added:  https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/authorized_key.py#L564
So there is not a good way to add comments, such as information about security or management policies, into the authorized key file.

EinYakAmNil:
##### SUMMARY

Fixes #256.
Added a new option "file_comment" to the module authorized_keys.
The feature would be nice to have and nobody else seems to be working on it.
I made the serialize function take an additional input parameter for the file comment.

Replacing the comment only works with exclusive=True.

##### ISSUE TYPE

- Feature Pull Request
-
##### COMPONENT NAME

authorized_keys

##### ADDITIONAL INFORMATION

The example should produce this

```paste below
# This file is managed by Ansible.
# All changes are temporary!
... some pubkeys
...
...
```


EinYakAmNil:
Hi maintainers, as far as  I can see, the failing checks are the result of the way the code was written before the CI pipeline existed.

Should the whole module be refactored, so the checks pass?
Would you want the refactor in this PR or a separate one?

